PR #607, opened 2014-12-13
Second submit newsletter

Christine612:
https://www.pivotaltracker.com/story/show/82978346
I found there is no change in my first submission. So I resubmit it. 
I have do some modification following your suggestions. I check the initialized variables, and the updates values after calling the method. Because I couldn't call the function explicitly, I just try to set the mode to instant, and do_send to be true, which might send newsletters for me automatically. However, I can only check the last_user_id is exactly the last person's id. I set the CHUNK_SIZE 100, but I cannot prove any difference, if there are more people than 100. I really appreciate your review and help.


yggie:
@Christine612 do you plan to continue work on this PR?

tochman:
I'm doing some spring cleaning and will close this one for now. Feel free to reopen if/when you feel like making contributions again.

PR #608, opened 2014-12-15
Image url

sidhene:
Created a field in the database, associated forms and views, spec tests, and validator in lib/validators that allow a user to include an image with their project description by linking to a file hosted on a whitelisted website.

https://www.pivotaltracker.com/story/show/80235250


sidhene:
I got distracted by the holiday madness. I'm about halfway through the changes and expect to finish up today.

spec/lib/validators/image_url_validator_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      validates_with ImageUrlValidator
+    end
+  end
+
+  subject { dummy_class.new }
+
+  it 'should be valid for a valid image url' do
+    subject.image_url = 'https://www.facebook.com/filename.jpg'
+    expect(subject).to be_valid
+  end
```
sidhene: Good to know.

sidhene:
Okay, I'm going to update and push, here is where we stand. Change requests sorted into done/not done, then in the order in which they appear on this page. 

DONE
http -> https
commented project.rb
whitespace
format divs
col-xs-0?
I don’t think this div tag has been closed properly
I don’t think the container-fluid class is necessary here
the indentation on this line is a little too large
v1 name is breaking Sephora
could this line be broken down into multiple lines?
 this regex match could be refactored into a separate method
 inconsistencies in the new lines used,
is_domain_whitelisted?
unclosed div tag?

NOT DONE
shoulda-matchers gem 
- - - skipped bc seems unimportant and I would need to pair to do it

meta-programming (shoulda-matchers gem)
- - - I can’t do this on my own and need to pair

asserts that this validator is being used for the Project model. 
- - I don’t understand what you mean here, but If you tell me which other file in our codebase does this, I should be able to follow that example. Also: I don’t see this assertion in any of the other validator specs, which suggests it’s a new standard and they need to be refactored to include it. 

whitelist method insecure
- True, but it was the best Christopher and I could come up with. If you have something better, paste the code snippet under your comment and I'll swap it out.

app/views/projects/_listing.html.erb
```diff
@@ -1,33 +1,45 @@
...
+    </div>
+    <div style="display: inline-block; font-size: 1.0em; font-style: italic;">
+      <div>Status: <strong style="font-style: normal;"><%= project.status.upcase %></strong></div>
+      <div><%= "Created #{time_ago_in_words(project.created_at)} ago" %> by <%= link_to project.user.display_name, user_path(project.user) %></div>
+    </div>
+</div>
+<div class="container-fluid">
+  <div class="row"> <div class="col-xs-0"></div><div class="col-xs-12">
+    <p><%= project.description.truncate(250, separator: ' ') %></p>
+  </div></div>
```
sidhene: I've corrected the formatting, and also double checked every div was closed in show, index, and _form.

sidhene:
Pushed. I think I can't submit another request until this one is closed

tochman:
I restarted semaphore build after initial failure (the Disqus thing again) All test are passing now. Good job.

My concern atm is again the commit history. I don't know if you did a `merge`, a `pull`or a `rebase` when doing https://github.com/sidhene/WebsiteOne/commit/51a87f234a5f36cf769bfa8f42aed131db3d7851 but something is not right.

tochman:
Also, since this is touching upon a lot of UI elements, is there a possibility of a demo deployment to Heroku so we can see this code live?

yggie:
@tochman I deployed this locally and it looks acceptable, we have a separate story for the UI: https://www.pivotaltracker.com/story/show/85684660. As for the commit history, it is a non-blocking issue, since there aren’t any other pressing matters, I will be merging this within the next hour (unless someone has any objections?)

tochman:
@yggie I'll take your word for it and yes, the commit history is not a blocker. Go ahead.

sidhene:
Thanks everyone :)

PR #617, opened 2015-01-12
Advanced user search2

christophermlne:
PR status: Advanced search features are working.
— needs some front-end love
— using ajax now to fetch search results (previous attempt without AJAX was too clunky)

Pivotal Story: https://www.pivotaltracker.com/story/show/78088460
Scratch Pad: https://docs.google.com/document/d/1zGiWLmGtaVWtt4lifgAgEZD-Uwy-2IhixnMIsoKzxko/edit?usp=sharing

### DIFF
--- a/app/assets/stylesheets/global/users.css.scss
+++ b/app/assets/stylesheets/global/users.css.scss
@@ -19,11 +19,7 @@
   }
 }
 
-.users-filter {
-  @extend .pull-right;
-  margin-top: 36px;
-}
-
+.users-filter { }
 
 .user-cancel-account {
   margin-top: 50px;
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,6 +6,7 @@ class UsersController < ApplicationController
 
   def index
     @users = User.search(params)
+    @projects = Project.all
   end
 
   def new
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,8 +5,7 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!, only: [:add_status]
 
   def index
-    count_timezone_degrees
-    @users = User.filter(params.slice(:project_filter, :timezone_filter, :online)).allow_to_display.by_create
+    @users = User.filter(set_filter_params).allow_to_display.by_create
     @users_count = User.allow_to_display.count
     @projects = Project.all
   end
@@ -69,15 +68,20 @@ class UsersController < ApplicationController
     @user = User.friendly.find(params[:id])
   end
 
-  def count_timezone_degrees
-    unless params[:timezone_select].blank?
-      if @current_user && @current_user.longitude
-        lng = @current_user.longitude
-        case params[:timezone_select]
+  def set_filter_params
+    filter_params = params.slice(:project_filter, :timezone_filter, :online)
+    set_timezone_degree_range(filter_params)
+    filter_params
+  end
+
+  def set_timezone_degree_range(filter_params)
+    unless filter_params[:timezone_filter].blank?
+      if lng = @current_user.try(:longitude)
+        case filter_params[:timezone_filter]
         when 'Close To My Timezone Area'
-          params[:timezone_filter] = [lng - 7.5, lng + 7.5]
+          filter_params[:timezone_filter] = [lng - 7.5, lng + 7.5]
         when 'Wider Timezone Area'
-          params[:timezone_filter] = [lng - 22.5, lng + 22.5]
+          filter_params[:timezone_filter] = [lng - 22.5, lng + 22.5]
         end
       else
         redirect_to :back, alert: "Can't determine your location!"
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,8 @@
 class UsersController < ApplicationController
   layout 'layouts/user_profile_layout', only: [:show]
 
+  skip_before_filter :verify_authenticity_token, :only => [:index, :show]
+
   before_filter :get_user, only: [:show, :add_status]
   before_filter :authenticate_user!, only: [:add_status]
 
@@ -8,6 +10,11 @@ class UsersController < ApplicationController
     @users = User.filter(set_filter_params).allow_to_display.by_create
     @users_count = User.allow_to_display.count
     @projects = Project.all
+
+    respond_to do |format|
+      format.js
+      format.html
+    end
   end
 
   def new
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -77,18 +77,18 @@ class UsersController < ApplicationController
 
   def set_filter_params
     filter_params = params.slice(:project_filter, :timezone_filter, :online)
-    set_timezone_degree_range(filter_params)
+    set_timezone_offset_range(filter_params)
     filter_params
   end
 
-  def set_timezone_degree_range(filter_params)
+  def set_timezone_offset_range(filter_params)
     unless filter_params[:timezone_filter].blank?
-      if lng = @current_user.try(:longitude)
+      if offset = @current_user.try(:timezone_offset)
         case filter_params[:timezone_filter]
-        when 'Close To My Timezone Area'
-          filter_params[:timezone_filter] = [lng - 7.5, lng + 7.5]
-        when 'Wider Timezone Area'
-          filter_params[:timezone_filter] = [lng - 22.5, lng + 22.5]
+        when 'In My Timezone'
+          filter_params[:timezone_filter] = [offset, offset]
+        when 'Members Within 2 Timezones'
+          filter_params[:timezone_filter] = [offset - 3600, offset + 3600]
         end
       else
         redirect_to :back, alert: "Can't determine your location!"
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,7 +5,8 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!, only: [:add_status]
 
   def index
-    @users = User.search(params)
+    @users = User.filter(params.slice(:project_filter)).allow_to_display.by_create
+    @users_count = User.allow_to_display.count
     @projects = Project.all
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,7 +5,8 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!, only: [:add_status]
 
   def index
-    @users = User.filter(params.slice(:project_filter)).allow_to_display.by_create
+    count_timezone_degrees
+    @users = User.filter(params.slice(:project_filter, :timezone_filter)).allow_to_display.by_create
     @users_count = User.allow_to_display.count
     @projects = Project.all
   end
@@ -66,5 +67,21 @@ class UsersController < ApplicationController
 
   def get_user
     @user = User.friendly.find(params[:id])
+  end
+
+  def count_timezone_degrees
+    unless params[:timezone_select].blank?
+      if @current_user && @current_user.longitude
+        lng = @current_user.longitude
+        case params[:timezone_select]
+        when 'Close To My Timezone Area'
+          params[:timezone_filter] = [lng - 7.5, lng + 7.5]
+        when 'Wider Timezone Area'
+          params[:timezone_filter] = [lng - 22.5, lng + 22.5]
+        end
+      else
+        redirect_to :back, alert: "Can't determine your location!"
+      end
+    end
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,7 +6,7 @@ class UsersController < ApplicationController
 
   def index
     count_timezone_degrees
-    @users = User.filter(params.slice(:project_filter, :timezone_filter)).allow_to_display.by_create
+    @users = User.filter(params.slice(:project_filter, :timezone_filter, :online)).allow_to_display.by_create
     @users_count = User.allow_to_display.count
     @projects = Project.all
   end
--- a/app/models/concerns/filterable.rb
+++ b/app/models/concerns/filterable.rb
@@ -1,0 +1,13 @@
+module Filterable
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def filter(filtering_params)
+      results = self.where(nil)
+      filtering_params.each do |key, value|
+        results = results.public_send(key, value) if value.present?
+      end
+      results
+    end
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -89,13 +89,12 @@ class User < ActiveRecord::Base
   end
 
   def self.search(params)
-    # TODO this code is not tested
     # NOTE maybe we create a filters hash ie., { filters: { project_filter: 6, another_filter: "foo"}
     # to make it easier to detect if a given search has filtering parameters set?
-    if params['project_filter']
+    if !params['project_filter'].blank?
       Project.find(params['project_filter']).followers
     else
-      # NOTE maybe we return a paginated list when no filtering criteria?
+      # NOTE return paginated list when no filtering criteria?
       where(display_profile: true)
         .order(:created_at)
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -53,6 +53,7 @@ class User < ActiveRecord::Base
   }
   scope :allow_to_display, -> { where(display_profile: true) }
   scope :by_create, -> { order(:created_at) }
+  scope :online, -> (argument) { where("updated_at > ?", 10.minutes.ago) }
 
   self.per_page = 30
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -87,10 +87,18 @@ class User < ActiveRecord::Base
   def is_privileged?
     Settings.privileged_users.split(',').include?(email)
   end
-  
+
   def self.search(params)
-    where(display_profile: true)
-      .order(:created_at)
+    # TODO this code is not tested
+    # NOTE maybe we create a filters hash ie., { filters: { project_filter: 6, another_filter: "foo"}
+    # to make it easier to detect if a given search has filtering parameters set?
+    if params['project_filter']
+      Project.find(params['project_filter']).followers
+    else
+      # NOTE maybe we return a paginated list when no filtering criteria?
+      where(display_profile: true)
+        .order(:created_at)
+    end
   end
 
   def self.find_by_github_username(username)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -48,6 +48,9 @@ class User < ActiveRecord::Base
       }
     )
   }
+  scope :timezone_filter, -> (degrees) {
+    where("longitude BETWEEN ? AND ?", degrees[0], degrees[1])
+  }
   scope :allow_to_display, -> { where(display_profile: true) }
   scope :by_create, -> { order(:created_at) }
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,11 +49,11 @@ class User < ActiveRecord::Base
     )
   }
   scope :timezone_filter, -> (degrees) {
-    where("longitude BETWEEN ? AND ?", degrees[0], degrees[1])
+    where("users.longitude BETWEEN ? AND ?", degrees[0], degrees[1])
   }
   scope :allow_to_display, -> { where(display_profile: true) }
   scope :by_create, -> { order(:created_at) }
-  scope :online, -> (argument) { where("updated_at > ?", 10.minutes.ago) }
+  scope :online, -> (argument) { where("users.updated_at > ?", 10.minutes.ago) }
 
   self.per_page = 30
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,6 @@
 class User < ActiveRecord::Base
+  include Filterable
+
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
@@ -34,7 +36,19 @@ class User < ActiveRecord::Base
   has_many :status
 
   accepts_nested_attributes_for :status
-  scope     :mail_receiver, -> { where(receive_mailings: true) }
+  scope :mail_receiver, -> { where(receive_mailings: true) }
+  scope :project_filter, -> (project_id) {
+    joins(:follows)
+    .where(
+      "\"follows\".\"blocked\" = 'f' 
+      AND \"follows\".\"followable_id\" = ? 
+      AND \"follows\".\"followable_type\" = 'Project' 
+      AND \"follows\".\"follower_type\" = 'User'",
+      project_id
+    )
+  }
+  scope :allow_to_display, -> { where(display_profile: true) }
+  scope :by_create, -> { order(:created_at) }
 
   self.per_page = 30
 
@@ -86,18 +100,6 @@ class User < ActiveRecord::Base
 
   def is_privileged?
     Settings.privileged_users.split(',').include?(email)
-  end
-
-  def self.search(params)
-    # NOTE maybe we create a filters hash ie., { filters: { project_filter: 6, another_filter: "foo"}
-    # to make it easier to detect if a given search has filtering parameters set?
-    if !params['project_filter'].blank?
-      Project.find(params['project_filter']).followers
-    else
-      # NOTE return paginated list when no filtering criteria?
-      where(display_profile: true)
-        .order(:created_at)
-    end
   end
 
   def self.find_by_github_username(username)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -40,11 +40,12 @@ class User < ActiveRecord::Base
   scope :project_filter, -> (project_id) {
     joins(:follows)
     .where(
-      "\"follows\".\"blocked\" = 'f' 
-      AND \"follows\".\"followable_id\" = ? 
-      AND \"follows\".\"followable_type\" = 'Project' 
-      AND \"follows\".\"follower_type\" = 'User'",
-      project_id
+      follows: {
+        blocked: false,
+        followable_id: project_id,
+        followable_type: 'Project',
+        follower_type: 'User'
+      }
     )
   }
   scope :allow_to_display, -> { where(display_profile: true) }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,6 +23,8 @@ class User < ActiveRecord::Base
   extend FriendlyId
   friendly_id :display_name, use: :slugged
 
+  before_save :generate_timezone_offset
+
   after_validation :geocode, if: ->(obj){ obj.last_sign_in_ip_changed? }
   after_validation -> { KarmaCalculator.new(self).perform }
   after_create :send_slack_invite, if: -> { Features.slack.invites.enabled }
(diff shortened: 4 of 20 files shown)


app/models/concerns/filterable.rb
```diff
@@ -0,0 +1,13 @@
+module Filterable
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def filter(filtering_params)
+      results = self.where(nil)
```
christophermlne: @Sysqa what is this line doing?

Sysqa: @christophermlne as I see in the command line this is equal to `self.all`. I thought on it also but I left it as I found.

christophermlne: so if i understand this method correctly, the way it works is results is initialized as all the records in the db for the model class, then within the loop a symbol representing a class method and a value representing an argument to that method are sent to the class the module is included on, causing the invocation of that method? why not use scopes or convert this method to a scope?

christophermlne: oh i  see now. :timezone_filter and :project_filter are scopes already

Sysqa: This method would be simplify the searching method and this method chain together the scopes in the the User model http://www.justinweiss.com/blog/2014/02/17/search-and-filter-rails-models-without-bloating-your-controller/ Yes it is chainable as you can see in the user controller https://github.com/AgileVentures/WebsiteOne/pull/617/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R8

christophermlne: it's great i think. took a bit of effort to understand it but i see how this will allow us to add new filtering options without increasing code complexity :+1:

app/controllers/users_controller.rb
```diff
@@ -64,4 +67,20 @@ def should_display_user?(user)
   def get_user
     @user = User.friendly.find(params[:id])
   end
+
+  def count_timezone_degrees
+    unless params[:timezone_select].blank?
+      if @current_user && @current_user.longitude
+        lng = @current_user.longitude
```
christophermlne: could lines 73 and 74 could be replaced with
```
if lng = @current_user.longitude
```

app/controllers/users_controller.rb
```diff
@@ -64,4 +67,20 @@ def should_display_user?(user)
   def get_user
     @user = User.friendly.find(params[:id])
   end
+
+  def count_timezone_degrees
+    unless params[:timezone_select].blank?
```
christophermlne: do we need line 72? 

looks like blank values are already guarded against in .filter

app/controllers/users_controller.rb
Sysqa: :+1:

app/controllers/users_controller.rb
Sysqa: yes but `params[:timezone_select]` is hold the selected option and filter get the `params[:timezone_filter]` what is different. It's a little bit confusing maybe, but I didn't want to overwrite what selected the user.

app/controllers/users_controller.rb
```diff
@@ -64,4 +67,20 @@ def should_display_user?(user)
   def get_user
     @user = User.friendly.find(params[:id])
   end
+
+  def count_timezone_degrees
```
christophermlne: rename something like 'set_timezone_filter_range' ?

Sysqa: :+1:  ok your name would be better

app/controllers/users_controller.rb
```diff
@@ -64,4 +67,20 @@ def should_display_user?(user)
...
+      if @current_user && @current_user.longitude
+        lng = @current_user.longitude
+        case params[:timezone_select]
+        when 'Close To My Timezone Area'
+          params[:timezone_filter] = [lng - 7.5, lng + 7.5]
+        when 'Wider Timezone Area'
+          params[:timezone_filter] = [lng - 22.5, lng + 22.5]
+        end
+      else
+        redirect_to :back, alert: "Can't determine your location!"
```
christophermlne: i wonder if we should disable the geolocation-based filter controls if there is no longitude value for the current user... with a label or tooltip to explain the reason why the controls are disabled... perhaps we tackle this once the feature is further along and we are in a position to consider user experience...

Sysqa: good idea

app/controllers/users_controller.rb
YaroSpace: if it is possible that `@current_user` maybe nil, then `lng = @current_user.try(:longitude)`

features/step_definitions/user_steps.rb
```diff
@@ -227,6 +227,23 @@
   end
 end
 
+Given /^the following active users exist$/ do |table|
+  table.hashes.each do |attributes|
+    p = Project.find_by(title: attributes['projects'])
+    u = FactoryGirl.create(
```
YaroSpace: expanding p and u into project and user would make it a little prettier :-)

features/step_definitions/user_steps.rb
```diff
@@ -391,3 +408,25 @@
...
+When(/^I select "(.*?)" from the "(.*?)" list$/) do |selected_from_list, list_name|
+  case list_name
+  when 'projects'
+    filter = 'project_filter'
+  when 'timezones'
+    filter = 'timezone_select'
+  when 'online status'
+    filter = 'online'
+  end
+
```
YaroSpace: can write
```
filter = case list_name
when 'projects' then 'project_filter'
....
end
```

features/users/user_list.feature
```diff
@@ -2,26 +2,36 @@
...
+      | hello world   | greetings earthlings    |             | active   | https://github.com/AgileVentures/WebsiteOne | https://www.pivotaltracker.com/s/projects/742821 |
+      | hello mars    | greetings aliens        |             | inactive |                                             |                                                  |
+      | hello jupiter | greetings jupiter folks |             | active   |                                             |                                                  |
+      | hello mercury | greetings mercury folks |             | inactive |                                             |                                                  |
+      | hello saturn  | greetings saturn folks  | My pitch... | active   |                                             |                                                  |
+      | hello sun     | greetings sun folks     |             | active   |                                             |                                                  |
+    And there are no videos
```
YaroSpace: for readability, I would include only the columns that are used in the scenarios.

features/users/user_list.feature
```diff
@@ -30,6 +40,37 @@ Feature: As a site owner
...
+    And I should not see "Dave"
 
+  Scenario: Filtering by project involvement
+    Given I am on the "our members" page
+    When I filter "projects" for "hello world"
+    Then I should see "Alice"
+    And I should see "Bob"
+    And I should not see "Croutch"
+    And I should not see "Dave"
+
```
YaroSpace: there is a table style step for I should/should not see:
```
I should see:
| Text |
| Alice |
| Bob |
```

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+      subject { User.filter(params.slice('project_filter')).allow_to_display }
+
+      before(:each) do
+        FactoryGirl.create(:user, first_name: 'Bob', created_at: 5.days.ago)
+        FactoryGirl.create(:user, first_name: 'Marley', created_at: 2.days.ago)
+        FactoryGirl.create(:user, first_name: 'Janice', display_profile: false)
+      end
+
+      it 'should be ordered by creation date' do
+        expect(subject.first.first_name).to eq('Bob')
```
YaroSpace: the creator of rspec suggests not to use 'should' in favour of 'it orders by creation date'.  He mentions that he used should when introducing rspec to make people think the correct way.

app/models/user.rb
```diff
@@ -34,7 +36,24 @@ class User < ActiveRecord::Base
...
+        follower_type: 'User'
+      }
+    )
+  }
+  scope :timezone_filter, -> (degrees) {
+    where("longitude BETWEEN ? AND ?", degrees[0], degrees[1])
+  }
+  scope :allow_to_display, -> { where(display_profile: true) }
+  scope :by_create, -> { order(:created_at) }
+  scope :online, -> (argument) { where("updated_at > ?", 10.minutes.ago) }
```
YaroSpace: I think all the scopes should be given names that make it obvious that they are scopes and not some method, e.g. name them all `by_....`

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+
+      context 'filters users for timezone area' do
+        before(:each) do
+          @current_user = FactoryGirl.create(:user, longitude: 19.15)
+        end
+        
+        it 'filters user1 when choose Close To My Timezone Area' do
+          params['timezone_filter'] = [@current_user.longitude - 7.5, @current_user.longitude + 7.5]
+
+          results = User.filter(params.slice('timezone_filter')).allow_to_display
```
YaroSpace: why bother using params if you can write directly
```
timezone_filter = ...
User.filter(timezone_filter)
```

app/controllers/users_controller.rb
```diff
@@ -4,7 +4,10 @@ class UsersController < ApplicationController
   before_filter :authenticate_user!, only: [:add_status]
 
   def index
-    @users = User.search(params)
+    count_timezone_degrees
+    @users = User.filter(params.slice(:project_filter, :timezone_filter, :online)).allow_to_display.by_create
```
YaroSpace: it is generally not a good idea to mutate a data structure, params hash in this case, from within somewhere else, count_timezone_degrees method.
I would prefer to have something like:
`tz_filter = set_time_zone_filter` and then pass it explicitly to User.filter

another concern is passing user submitted params directly to scopes.  I wonder if there is a security problem here.

YaroSpace:
Great job guys.  Do not mind y opinionated comments :-)

Sysqa:
Thanks @YaroSpace, I think we can evolve if somebody give us good ideas. :)

features/users/user_list.feature
christophermlne: :+1:

tochman:
@Sysqa @christophermlne Any action on this PR. Do you want to pair up and make a push together?

Sysqa:
@tochman sure we can try to do something at weekend. @christophermlne so you have time to do pairing somewhat?

christophermlne:
hi yes. i can devote some time this weekend to this PR.

are you around tomorrow? i think we're nearly done. some minor cleanup and front-end stuff...

Sysqa:
Sure, but I'm afraid I'll have time till 7pm in my time, in your time about 1pm

tochman:
Good work guys! I'll try to review this ASAP (or get @yggie to do it) so we can merge it in.

app/assets/stylesheets/global/users.css.scss
```diff
@@ -19,11 +19,7 @@
   }
 }
 
-.users-filter {
-  @extend .pull-right;
-  margin-top: 36px;
-}
-
+.users-filter { }
```
yggie: does this do anything?

app/controllers/users_controller.rb
```diff
@@ -1,11 +1,20 @@
...
   def index
-    @users = User.search(params)
+    @users = User.filter(set_filter_params).allow_to_display.by_create
+    @users_count = User.allow_to_display.count
+    @projects = Project.all
+
+    respond_to do |format|
+      format.js
+      format.html
+    end
```
yggie: do we need the `respond_to` block here?

yggie: ok nevermind, I see you using it later in the PR :sweat_smile:

yggie: hmm… seem to be getting an error response when I use the “Close To My Timezone Area” option:

```
AbstractController::DoubleRenderError at /users
===============================================

> Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".

app/controllers/users_controller.rb, line 14
--------------------------------------------

    9     def index
   10       @users = User.filter(set_filter_params).allow_to_display.by_create
   11       @users_count = User.allow_to_display.count
   12       @projects = Project.all
   13   
>  14       respond_to do |format|
   15         format.js
   16         format.html
   17       end
   18     end
   19   

App backtrace
-------------

 - app/controllers/users_controller.rb:14:in `index'
```

could you look into this?

app/controllers/users_controller.rb
```diff
@@ -65,4 +74,25 @@ def should_display_user?(user)
...
+    unless filter_params[:timezone_filter].blank?
+      if lng = @current_user.try(:longitude)
+        case filter_params[:timezone_filter]
+        when 'Close To My Timezone Area'
+          filter_params[:timezone_filter] = [lng - 7.5, lng + 7.5]
+        when 'Wider Timezone Area'
+          filter_params[:timezone_filter] = [lng - 22.5, lng + 22.5]
+        end
+      else
+        redirect_to :back, alert: "Can't determine your location!"
```
yggie: with regard to my previous comment, my guess is that this is the issue. `render` is still being called regardless of whether the `redirect` takes place. Could the `redirect` logic be moved inside the controller method or have the private methods renamed to indicate that a redirect might take place? It is hard to see where the response goes if the private methods are hiding a `redirect`.

app/controllers/users_controller.rb
```diff
@@ -65,4 +74,25 @@ def should_display_user?(user)
   def get_user
     @user = User.friendly.find(params[:id])
   end
+
+  def set_filter_params
+    filter_params = params.slice(:project_filter, :timezone_filter, :online)
+    set_timezone_degree_range(filter_params)
+    filter_params
+  end
```
yggie: Also, it is not a good idea for a method called `set_xxx` to be redirecting anywhere :wink:, if the method is called `set_xxx`, it should do just that, set `xxx` to some new value.

On a different note, this method isn’t really just setting an instance variable (as one would expect from the method signature), it is really getting/computing the filter params. Perhaps it should just be called `filter_params`?

app/models/concerns/filterable.rb
```diff
@@ -0,0 +1,13 @@
...
+      filtering_params.each do |key, value|
+        results = results.public_send(key, value) if value.present?
+      end
```
yggie: I personally don’t think we should be accepting every key/value pair as a valid method/input. It does make the code DRY but it is hard predict the behaviour. Could we implement a whitelist perhaps to limit the accepted parameters?

app/models/user.rb
```diff
@@ -34,7 +36,24 @@ class User < ActiveRecord::Base
...
+        followable_id: project_id,
+        followable_type: 'Project',
+        follower_type: 'User'
+      }
+    )
+  }
+  scope :timezone_filter, -> (degrees) {
+    where("users.longitude BETWEEN ? AND ?", degrees[0], degrees[1])
+  }
+  scope :allow_to_display, -> { where(display_profile: true) }
```
yggie: perhaps this could be renamed to `public_accounts` or something similar?

app/models/user.rb
```diff
@@ -34,7 +36,24 @@ class User < ActiveRecord::Base
...
+  scope :by_create, -> { order(:created_at) }
```
yggie: `by_create` doesn’t really say much about what it is doing, a better name would be `order_by_created_at`, but at this point, it might just be better to use `order(:created_at)` rather than a scope!

app/models/user.rb
```diff
@@ -34,7 +36,24 @@ class User < ActiveRecord::Base
...
+  scope :online, -> (argument) { where("users.updated_at > ?", 10.minutes.ago) }
```
yggie: `10.minutes` is repeated here and in the `online?` method. Could we unify the behaviour somehow?

app/controllers/users_controller.rb
```diff
@@ -65,4 +74,25 @@ def should_display_user?(user)
...
+
+  def set_timezone_degree_range(filter_params)
+    unless filter_params[:timezone_filter].blank?
+      if lng = @current_user.try(:longitude)
+        case filter_params[:timezone_filter]
+        when 'Close To My Timezone Area'
+          filter_params[:timezone_filter] = [lng - 7.5, lng + 7.5]
+        when 'Wider Timezone Area'
+          filter_params[:timezone_filter] = [lng - 22.5, lng + 22.5]
```
yggie: do these numbers represent anything significant? (like time zone difference) or are they just arbitrary numbers?

app/views/users/index.js.erb
```diff
@@ -0,0 +1 @@
+$('#UsersList').html("<%= escape_javascript(render 'user_list') %>");
```
yggie: how would you trigger this response?

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+
+      it 'filters users for project' do
+        @user1.follow @project
+        @user2.stop_following @project
+        params['project_filter'] = @project.id
+
+        results = User.filter(params).allow_to_display
+        
+        expect(results).to include(@user1)
+        expect(results).not_to include(@user2)
```
yggie: you could use the `match_array` matcher for these expectations, that way you will only need to write one `expect`

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+        it 'filters user1 when choose Close To My Timezone Area' do
+          params['timezone_filter'] = [@current_user.longitude - 7.5, @current_user.longitude + 7.5]
+
+          results = User.filter(params).allow_to_display
+        
+          expect(results).to include(@user1)
+          expect(results).not_to include(@user2)
+        end
+
+        it 'filters both users when choose Wider Timezone Area' do
```
yggie: I don’t think this description isn’t quite right, you are simply testing the `timezone_filter` here. This description would be more suitable for a test in the `users_controller_spec`, where a possible input would actually be “Wider Timezone Area”. In here, you would be assuming a relationship between the `users_controller` and `User`, which is outside the scope of this test context.

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+        expect(results).to include(@user1)
+        expect(results).not_to include(@user2)
+      end
+
+      context 'filters users for timezone area' do
+        before(:each) do
+          @current_user = FactoryGirl.create(:user, longitude: 19.15)
+        end
+        
+        it 'filters user1 when choose Close To My Timezone Area' do
```
yggie: same here :grin:

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+
+      before(:each) do
+        FactoryGirl.create(:user, first_name: 'Bob', created_at: 5.days.ago)
+        FactoryGirl.create(:user, first_name: 'Marley', created_at: 2.days.ago)
+        FactoryGirl.create(:user, first_name: 'Janice', display_profile: false)
+      end
+
+      it 'ordered by creation date' do
+        expect(subject.first.first_name).to eq('Bob')
+      end
```
yggie: correct me if I am wrong, but I don’t believe I have seen anywhere which explicitly stated that the ordering would be by `created_at` date, I believe this is simply the default behaviour of `ActiveRecord`. If this is the case, then this test is not necessary :grinning:

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+          @current_user = FactoryGirl.create(:user, longitude: 19.15)
+        end
+        
+        it 'filters user1 when choose Close To My Timezone Area' do
+          params['timezone_filter'] = [@current_user.longitude - 7.5, @current_user.longitude + 7.5]
+
+          results = User.filter(params).allow_to_display
+        
+          expect(results).to include(@user1)
+          expect(results).not_to include(@user2)
```
yggie: :arrow_right: `match_array`

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+        end
+
+        it 'filters both users when choose Wider Timezone Area' do
+          params['timezone_filter'] = [@current_user.longitude - 22.5, @current_user.longitude + 22.5]
+
+          results = User.filter(params).allow_to_display
+        
+          expect(results).to include(@user1)
+          expect(results).to include(@user2)
```
yggie: `match_array` :arrow_left:

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+      end
 
-    it 'should be filtered by the display_profile property' do
-      results = subject.map(&:first_name)
-      expect(results).to include('Marley')
-      expect(results).not_to include('Janice')
+      it 'filtered by the display_profile property' do
+        results = subject.map(&:first_name)
+        expect(results).to include('Marley')
+        expect(results).not_to include('Janice')
```
yggie: :arrow_down: 
`match_array`

spec/models/user_spec.rb
```diff
@@ -176,24 +176,77 @@
...
+          expect(results).to include(@user2)
+        end
+      end
+
+      it 'does not raise error when filters are empty' do
+        params['project_filter'] = ''
+        params['timezone_filter'] = ''
+
+        expect{ User.filter(params).allow_to_display }.to_not raise_error
+      end
```
yggie: While I understand the importance of this test, perhaps it would be better to show the behaviour of the filters when the parameters are empty? (what results are returned, whether they make sense or not) Not raising an error tells me very little about it’s impact on the application.

spec/models/user_spec.rb
yggie: that being said, you are missing tests in `users_controller_spec`, especially for that bug I discovered :wink:

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
     <% else %>
-        <p>Check out our <%= @users.count %> awesome volunteers from all over the globe!</p>
+        <p>Check out our <%= @users_count %> awesome volunteers from all over the globe!</p>
     <% end %>
   </div>
-  <div class="col-sm-4">
+  <% if @current_user %>
+    <div class="col-sm-12 well">
+      <p>Filter users by</p>
+      <%= form_tag(users_path, remote: true, method: "get", class: "form-inline filters-users-advanced") do %>
```
yggie: any reason why this is an ajax request? the final behaviour acts pretty much like a normal request (but without the progress indicator)

app/views/users/index.js.erb
yggie: oh nevermind, figured it out

app/views/users/index.html.erb
yggie: Also, if the request isn’t done with ajax, you get the added benefit of keeping the URL stateful so refreshing the page won’t lose your results. You also get rid of the extra `.js` format option, so the controller will look a lot more cleaner, but I am curious to hear why you guys chose this route to begin with.

yggie:
sorry it took so long to have a look! been so busy lately :weary:

app/views/users/index.html.erb
tochman: I think it enhances the user experience to use ajax here and I was advising the use of it on this particular feature.

tochman:
@yggie @christophermlne I would say that we can move ahead with this PR in it's current state.

Sysqa:
@tochman as I see I have some modification what is missing from this branch, and yeah maybe it needs to modify what @yggie suggested.

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
+    <h1>Members Directory</h1>
+    <% if @users_count == 0 %>
         <p>It is a lonely planet we live in&#8230;</p>
     <% else %>
-        <p>Check out our <%= @users.count %> awesome volunteers from all over the globe!</p>
+        <p>Check out our <%= @users_count %> awesome volunteers from all over the globe!</p>
     <% end %>
   </div>
-  <div class="col-sm-4">
+  <% if @current_user %>
```
tochman: I would not limit the feature to signed in users only.

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
+    <div class="col-sm-12 well">
+      <p>Filter users by</p>
+      <%= form_tag(users_path, remote: true, method: "get", class: "form-inline filters-users-advanced") do %>
+        <%= select_tag :project_filter, options_from_collection_for_select(@projects, 'id', 'title', params[:project_filter]), {  prompt: 'Project involvement...' } %>
+        <%= select_tag :timezone_filter, options_for_select(['Close To My Timezone Area', 'Wider Timezone Area'], params[:timezone_filter]), { prompt: 'Time zone...' } %>
```
tochman: I would suggest that this field should be hidden if user is not geocoded or not signed in.

tochman: I would also add `{ prompt: 'Time zone...', class: 'form-control'}` to use bootstrap form class. On both `select_tag`'s

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
+        <%= check_box_tag :online, params[:online], (true if !params[:online].nil?) %> Recently Online
+        <%= submit_tag 'submit' %>
```
tochman: `<%= submit_tag 'Search', class: 'btn btn-default' %>` gives the button a bootstrap look.

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
+      <% end %>
+    </div>
+  <% end %>
```
tochman: All in all, I would suggest:
```
  <div class="col-sm-12 well">
    <p>Filter users by</p>
    <%= form_tag(users_path, remote: true, method: "get", class: "form-inline filters-users-advanced") do %>
        <%= select_tag :project_filter, options_from_collection_for_select(@projects, 'id', 'title', params[:project_filter]), {prompt: 'Project involvement...', class: 'form-control'} %>
        <% if @current_user.latitude %>
            <%= select_tag :timezone_filter, options_for_select(['Close To My Timezone Area', 'Wider Timezone Area'], params[:timezone_filter]), {prompt: 'Time zone...', class: 'form-control'} %>
        <% end %>
        <%= check_box_tag :online, params[:online], (true if !params[:online].nil?) %> Recently Online
        <%= submit_tag 'Search', class: 'btn btn-default' %>
    <% end %>
  </div>
```

app/views/users/index.html.erb
```diff
@@ -1,54 +1,42 @@
...
+        <input type="text" class="form-control" id="user-filter" placeholder="Search members...">
       </div>
     </form>
   </div>
 </div>
 <div class="row">
-  <% unless @users.empty? %>
+  <div class="col-sm-12">
+    <% unless @users.empty? %>
       <ul class="media-list row" id="UsersList">
```
tochman: Although there isn't a standard naming convention for html/css ids, I would set this one to `user-list` The most common naming standards are:
There are several options:
```
id="myId"
id="my-id"
id="my_id"
```
From all of these I think that the dash look a lot cleaner than under_scores or camelCase.

app/views/users/index.js.erb
tochman: I was about to suggest `<%= j render 'user_list' %>` as a cleaner way to do the same thing as `escape_javascript(...)` but that's just excessive and obnoxious. ;-)

tochman:
I deployed a demo with my proposed changes and @Sysqa 's timezone code to https://wso-map.herokuapp.com/users
skarmavbild 2015-03-11 kl 11 34 29

Sysqa:
Looks good, but as I seen there is an Internal Server Error for the non logged in user...

tochman:
True @Sysqa Fixed that on the demo site now. Updated the code snippet above as well. Good catch.

Sysqa:
:+1:

tochman:
Semaphore is not picking up our pull requests right now (I'm working with their support on it). But I'll merge this in for now and see if the build is picked up - we can always revert the merge if something is wrong.

PR #620, opened 2015-03-12
Tweet YT recordings

tochman:
PT story (addition to previous https://github.com/AgileVentures/WebsiteOne/pull/613 and https://github.com/AgileVentures/WebsiteOne/pull/592) https://www.pivotaltracker.com/story/show/65542324

- Triggers a tweet (TwitterService) with link to recording when `yt_video_id` is present on an EventInstance.
- One version for 'Scrum' and one version for 'PairProgramming' categories

### DIFF
--- a/app/models/event_instance.rb
+++ b/app/models/event_instance.rb
@@ -13,6 +13,7 @@ class EventInstance < ActiveRecord::Base
   scope :pp_hangouts, -> { where(category: 'PairProgramming') }
 
   after_save "tweet_hangout_notification if (started? && hangout_url_changed?)"
+  after_save "tweet_yt_link if yt_video_id_changed?"
 
   def started?
     hangout_url?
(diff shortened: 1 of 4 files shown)


tochman:
Closing and reopening due to Semaphore issues.

app/models/event_instance.rb
```diff
@@ -38,12 +39,25 @@ def start_datetime
...
+    case self.category
+      when 'Scrum'
+        TwitterService.tweet("#{broadcaster} just hosted an online #scrum using ##googlehangouts Missed it? Catch the recording at youtu.be/#{self.yt_video_id} #opensource")
+      when 'PairProgramming'
+        TwitterService.tweet("#{broadcaster} just finished #PairProgramming on #{self.project.title} You can catch the recording at youtu.be/#{self.yt_video_id} #opensource #pairwithme")
+    end
+  end
+
+  def broadcaster
+    self.participants.each { |_, hash| break hash['person']['displayName'] if hash['isBroadcaster'] == 'true' }
```
christophermlne: what does the underscore argument do?

sampritipanda: The underscore argument usually refers to a necessary block argument which is not used in the block.

tochman: To be honest? NO CLUE! ;-) Got this method from SO See: http://stackoverflow.com/questions/29019041/object-from-nested-hash-depending-on-key-value

christophermlne: found a write up about it here: http://po-ru.com/diary/rubys-magic-underscore/

@sampritipanda is correct.

tochman: Good post, now I can actually understand what I'm doing (in part, haha) Anyway do upvote the accepted answer on SO if you have time. It saved me a lot of time.

app/models/event_instance.rb
```diff
@@ -38,12 +39,25 @@ def start_datetime
...
+        TwitterService.tweet("Pair programming on #{self.project.title} #{hangout_url} #pairwithme, #distributedteam")
+      else
+        Rails.logger.error "''#{self.category}'' is not a recognized event category for Twitter notifications. Must be one of: 'Scrum', 'PairProgramming'"
     end
   end
+
+  def tweet_yt_link
+    case self.category
+      when 'Scrum'
+        TwitterService.tweet("#{broadcaster} just hosted an online #scrum using ##googlehangouts Missed it? Catch the recording at youtu.be/#{self.yt_video_id} #opensource")
```
christophermlne: i'm wondering is the twitter message templates should move to a yaml file or are by some other means separated from the model? phase 3 maybe?

tochman: Sure. Good suggestion.

PR #636, opened 2015-03-21
Patch - Change "Close To My Timezone Area" to "In My Timezone"

Sysqa:
As @tochman asked I change `Close To My Timezone Area` to `In my timezone`


christophermlne:
should it be 'In or near my timezone'?

tochman:
Well, is it 'in' or 'in and near'? What results is this search returning?  When I tried it it retured only members IN my timezone.

tochman:
@Sysqa you need to merge with upstream. Your ruby version is not correct. We bumbed it up to 2.1.5

christophermlne:
i believe it is near. if you live near the edge of your timezone you will get members from an adjacent timezone

Sysqa:
After timezone offset modifications the In My Timezone should be the right...

PR #641, opened 2015-03-30
35px avatar on navbar

achock:
https://www.pivotaltracker.com/n/projects/982890/stories/76848756


tochman:
This looks great but we need a test for this change.

yggie:
@tochman I don’t think it is necessary to test for this change, UI shouldn’t be too coupled to the tests, it makes it much more brittle. In my opinion, this is good to go

tochman:
Good point @yggie

PR #657, opened 2015-04-03
Fix n+1 issue on users#index:

alejandrobabio:
* Eager load of association on users: ':status, :titles, :skills'.
* Choose '.size' instead of '.count' on associations.
* Do not use <tag>_list (acts-as-taggable-on method). Instead map data
*  from cached association.


alejandrobabio:
I close because I need do a revision.

PR #659, opened 2015-04-06
WIP: Add advanced search

msausa:
The search bar is not rendering on any page other than the home page. Raoul has discussed with Thomas who believes he knows what the issue is (the problem is related to turbolinks). Work in progress.


diraulo:
Published on heroku ==> https://wso-search.herokuapp.com/

tochman:
You have to run `heroku run rake db:seed` so there is something to search for.

diraulo:
@tochman done.

alejandrobabio:
I realise that the google search send me to the **production** AgileVentures site. And that's why the search icon is not there.

tochman:
Yeah, that a fact. But the problem is that when you load the app you get the icon and the search bar:
skarmavbild 2015-04-07 kl 15 32 02
Then you navigate to some other page and you get this:
skarmavbild 2015-04-07 kl 15 31 47
And then when you reload that page, you get this:
skarmavbild 2015-04-07 kl 15 32 21
So it is a turbolinks issue. The javascript is not being loaded on document ready. This is a site wide issue and we have faced it before. I don't recal how we fixed it to be honest. There is a SO question that is worth exploring: http://stackoverflow.com/questions/18770517/rails-4-how-to-use-document-ready-with-turbo-links

tochman:
@sampritipanda @YaroSpace @yggie Can you have a look at this issue. I think you guys have some experience with this on WSO

alejandrobabio:
I think it can be fixed changing ```(function() { ... })();``` with ```$(document).on('ready page:load', function(e) { ... });``` on ```app/views/layouts/_navbar.html.erb```

diraulo:
We discussed this issue with @yggie yesterday and it seems the google script itself is using an approach that is incompatible with turbolinks. Bryan suggested we either stop using turbolinks in the application or we could implement a custom search for the application.

app/assets/javascripts/search_toggle.js
```diff
@@ -0,0 +1,16 @@
...
+        } else {
+            $('#google_search i').removeClass('fa-times').addClass('fa-search');
+        }
+
+        $('#google_search_wrapper').fadeToggle(200, function() {
+            $("#gsc-i-id1").focus();
+        });
+    });
+    return false;
+});
```
tochman: Okay, so I would like you to try this:

```
function initGoogleSearch() {
   $("#google_search").click(function() {
         // Replace search icon with close
         if($('#google_search_wrapper').css('display') == "none") {
             $('#google_search i').removeClass('fa-search').addClass('fa-times');
             $('#nav').css('z-index', 1000);
         } else {
             $('#google_search i').removeClass('fa-times').addClass('fa-search');
         }
 
         $('#google_search_wrapper').fadeToggle(200, function() {
             $("#gsc-i-id1").focus();
         });
     });
     return false;
}

$(document).ready(function () {
    initGoogleSearch()
});
document.addEventListener('page:load', function () {
   initGoogleSearch()
});
```

Have not tested it, but... yeah, le's see if it works

diraulo:
@tochman this block of js code is only responsible for toggling the search form visibility. As far as i can tell this works fine as it is. The issue we have is with reloading the actual snippet code from Google which currently in `_navbar.html.erb`
We extracted the following
```html
<script>
  (function() {
    var cx = '013004698248256241028:fzuo22ktbbo';
    var gcse = document.createElement('script');
    gcse.type = 'text/javascript';
    gcse.async = true;
    gcse.src = (document.location.protocol == 'https:' ? 'https:' : 'http:') +
        '//www.google.com/cse/cse.js?cx=' + cx;
    var s = document.getElementsByTagName('script')[0];
    s.parentNode.insertBefore(gcse, s);
  })();
</script>

```
to this:
```html
<script type="text/javascript" src="//www.google.com/cse/cse.js?cx=013004698248256241028:fzuo22ktbbo" data-turbolinks-track=false></script>
```

then moved it to the `head` section and still we had no luck.

tochman:
Okay, I see you point. Let's try to pair on this soon. Do you plan to host a session tonight?

diraulo:
Most probably, i was looking into other stuff to do while someone potentially finds a solution to this :smile:  I tried everything i could think of so far.

diraulo:
Check updated build here ==> https://wso-search.herokuapp.com/

PR #665, opened 2015-04-11
User counter cache

alejandrobabio:
Add counter cache to User for status association.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -10,6 +10,8 @@ describe User, :type => :model do
 
   it { is_expected.to accept_nested_attributes_for :status }
 
+  it { is_expected.to respond_to :status_count }
+
   it 'should have valid factory' do
     expect(FactoryGirl.create(:user)).to be_valid
   end
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ config/environments/*.local.yml
 #
 # Guard config
 Guardfile
+
+.vagrant/
+tags
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -1,5 +1,5 @@
 class Status < ActiveRecord::Base
-  belongs_to :user
+  belongs_to :user, counter_cache: :status_count
 
 
   OPTIONS = ['Ready to pair',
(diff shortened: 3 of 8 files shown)


alejandrobabio:
FYI: Task at PT https://www.pivotaltracker.com/story/show/80164906

spec/models/user_spec.rb
```diff
@@ -10,6 +10,8 @@
 
   it { is_expected.to accept_nested_attributes_for :status }
 
+  it { is_expected.to respond_to :status_count }
```
yggie: I think it would be better to test what `status_count` is doing rather than just testing that it responds to the method

alejandrobabio:
I'm agree. I'll improve it and send a new PR.

alejandrobabio:
@yggie do you think is enough testing the counter cache is set at status model?
```ruby
  it { is_expected.to belong_to(:user).counter_cache(:status_count)}
```
And have confidence that Rails test its behavior. Considering ```status_count``` value is required in these tests: ```spec/presenters/users/user_presenter_spec.rb:85``` and ```spec/views/users/index.html.erb_spec.rb:107```

yggie:
@alejandrobabio that looks like a much better alternative, I wasn’t aware that there was a `counter_cache` matcher, that will work much better

tochman:
Getting this error when running your migration:
```
Running `rake db:migrate` attached to terminal... up, run.7052
Migrating to AddStatusCountToUsers (20150410173625)
== 20150410173625 AddStatusCountToUsers: migrating ============================
-- add_column(:users, :status_count, :integer, {:default=>0})
   -> 0.1908s
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

status_count is marked as readonly/app/vendor/bundle/ruby/2.1.0/gems/activerecord-4.1.0/lib/active_record/persistence.rb:510:in `verify_readonly_attribute'
```

alejandrobabio:
@tochman I'm not sure over what commit I should make the patch.

tochman:
You need to create and submit a new PR @alejandrobabio

PR #670, opened 2015-04-14
Updated therubyracer version

yggie:
- Updated therubyracer version, this will bump the version of libv8 up
  to make it more compatible with Mac OS X Yosemite

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -236,7 +236,7 @@ GEM
     kdtree (0.3)
     launchy (2.4.2)
       addressable (~> 2.3)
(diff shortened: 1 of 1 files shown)


yggie:
@tochman good point, since `libv8` wasn’t a direct dependency, I though I would update the `Gemfile.lock` instead of adding it to the `Gemfile`. I ended up just updating `therubyracer` (which also bumps the version of `libv8` up. `bundle` uses `Gemfile.lock` to manage the exact versions, so it isn’t an issue on development.

tochman:
Cool, good solution.

PR #674, opened 2015-04-18
add a comment that you need to join the project to see the hangout button

piotrkedzia:
I modified views/project/show and partial that render hangout button.
New hangout button for users that no follow project works great but many cucumber tests are now failing.
In my opinion these failing test are no related to project/show view.
So Thomas if you have a moment please take a look on this.

https://www.pivotaltracker.com/story/show/76848756

### DIFF
--- a/app/views/event_instances/_hangout_button.html.erb
+++ b/app/views/event_instances/_hangout_button.html.erb
@@ -1,4 +1,5 @@
-<% start_data = JSON.generate({
+<% if following_project %>
+  <% start_data = JSON.generate({
     'title' => title,
     'category' => category,
     'projectId' => project_id.to_s,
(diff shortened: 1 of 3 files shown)


tochman:
We have failing tests on this one and we also need to test for the new behavior.

app/views/event_instances/_hangout_button.html.erb
```diff
@@ -8,8 +9,32 @@
     'callbackUrl' => hangout_url('id').gsub(/id$/, '')
   }) %>
```
tochman: I would suggest moving this out to a helper class. It feels like too much logic for the view. It was a cornercase before and it definitely now with this if condition.

piotrkedzia:
Moved hangout button for no followers frow partial direct to project/show view.
Now tests should pass.

app/views/event_instances/_hangout_button.html.erb
piotrkedzia: Moved all changes from this partial to view. So this file is untouched by me right now.

piotrkedzia:
I'll add cucumber tests as soon as posible.

piotrkedzia:
Thomas: I added a test that we talked about during the Friday WSO meeting. I hope now it's ready to merge :)

tochman:
Cool. Great work.

PR #696, opened 2015-06-16
PATCH: Fix for 500 error /scrums

tochman:
#### PT bug: https://www.pivotaltracker.com/story/show/97093234

- removed deprecated YT API Gem
- app is fetching info about past scrums from db instead of making API calls
- refactored cukes and hooks related to scrums feature


alejandrobabio:
@tochman with this you are implementing the develop branch (modified). Also you are implementing my updates at my branch 'hangouts', which is not ready.

Please do ```git diff master``` to see it.

PR #709, opened 2015-07-03
Trace start and end of broadcasting at hangouts.

alejandrobabio:
* Add column :hoa_status to event_instances table.
* EventInstancesController#update receive hoa_status from
  HangoutConnection
* EventInstance model don't update after set hoa_status to finished
* EventInstance#live? respond true if not finished and updated within 2
  minutes.
* Remove 'request.scheme' from CallbackURL hangout button field, to make
  posible use with either http or https protocol.
* Set Settings.hangout.app_id and Settings.slack.notifications and
  Settings.twitter.notifications to ENV variables to prevent calls from
  development & staging.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -570,6 +570,3 @@ DEPENDENCIES
   will_paginate-bootstrap
   yui-compressor
   zeus
-
-BUNDLED WITH
-   1.10.3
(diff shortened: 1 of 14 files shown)


alejandrobabio:
Add to WSO Heroku the new env variables:
* Development:
 * ENV['HANGOUTS_APP_ID']= 331105831845

* Production:
 * ENV['HANGOUTS_APP_ID']= 135304697558
 * ENV['SLACK_NOTIFICATIONS_ENABLED']= true
 * ENV['TWITTER_NOTIFICATIONS_ENABLED'] = true

Now, we can move WSO and HangoutConnection safely to develop -it requires migration of db-, and to production.

PR #712, opened 2015-07-16
Use ssl url for communication with hangouts

alejandrobabio:
Set the ssl site at `ENV['SSL_HOST']` this value must me `websiteone-production.herokuapp.com`
Then send to hangouts the ssl site at `callbackUrl`, with this value if it is set, otherwise with `root_url`.

At this moment, some disqus test are failing. 

This PR alone will not have effect, until the hot fix done at HangoutConnection will be reverted.

### DIFF
--- a/app/views/event_instances/_hangout_button.html.erb
+++ b/app/views/event_instances/_hangout_button.html.erb
@@ -5,7 +5,7 @@
   'eventId' => event_id.to_s,
   'hostId' => current_user.id,
   'hangoutId' => event_instance_id.present? ? event_instance_id : generate_event_instance_id(current_user, project_id),
-  'callbackUrl' => hangout_url('id').gsub(/id$/, '').sub(/^#{request.scheme}:/, '')
+  'callbackUrl' => url_for(controller: :event_instances, action: :update, id: 'id', protocol: 'https', host: (Settings.hangouts.ssl_host || root_url[0..-2])).gsub(/id$/, '')
 }) %>
 
 <div id="liveHOA-placeholder"
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -43,6 +43,7 @@ disqus:
 
 hangouts:
   app_id: <%= ENV['HANGOUTS_APP_ID'] %>
+  ssl_host: <%= ENV['SSL_HOST'] %>
 
 twitter:
   consumer_key:        nil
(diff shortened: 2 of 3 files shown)


diraulo:
@alejandrobabio when you get a chance, get changes from `upstream` and update this PR. ;)

alejandrobabio:
@diraulo I did update this PR. As I said beflore, it can be moved, also it is required add the `SSL_HOST` env var. And it will not have effect until HangoutConnection hot fix will be reverted.

tansaku:
hi @alejandrobabio thanks so much for your ongoing efforts here.  Forgive me for not remembering.  What's the reason for this combined PR and HangoutConnection hot fix reversion?  Is it part of moving completely away from Yaro's personal accounts etc.?

I remember we discussed before the summer, and I made some changes to hangout connection, but will need your help to coordinate pulling this in :-)

alejandrobabio:
Hi @tansaku, there are several things related.

Currently HC calls WSO with a hardcoded url `https://websiteone-production.herokuapp.com/hangouts/`. It should be working with the url sent from WSO `http://www.agileventures.org/hangouts/` (built at the hangout button), but it does not work. The problem here is the `http` protocol, hangouts requires `https`. And, that's why the url is hardcoded at HC.

This PR is a workaround on this problem, adding the environment variable `SSL_HOST` we can let WSO decide which https site will be used at HC updates. And, of course, isolate HC of knowing the WSO url for updates.

Then at WSO development we can set it's value to 'websiteone-develop.herokuapp.com`.
At production we need to use `websiteone-production.herokuapp.com`. And all must work fine.

The matter here is that can't be complete tested until it be at production. Because production is the only one that uses `http` and `https` with different urls.

I think the steps needed here are:

1. Move this PR to develop, set `SSL_HOST` at develop site.
2. Revert HC hardcode url at develop
3. Test that all work fine, and without generating updates of hangout to production
4. Move this PR to production, set `SSL_HOST` at production.
*** here start the risk of a fail at production ***
5. Update HC production with develop branch
6. Test at production if all is working fine.

After, or before all this, we can change the `HANGOUTS_APP_ID`, stop using Yaro's app and use our clone of that.

Please, ask me whatever you need. And, if you want we can coordinate to do all this together.

tansaku:
@alejandrobabio thanks for the detailed explanation - this is the related ticket right?: https://www.pivotaltracker.com/story/show/91253176

@diraulo will push to develop - and coordinate with you to do the push to production

alejandrobabio:
@diraulo just let me know when heroku develop were updated with this PR, and set the var: `SSL_HOST`.

diraulo:
@alejandrobabio i merged this PR and it was published to develop 2 days ago. I've set the `SSL_HOST` to `https://websiteone-develop.herokuapp.com`

Ping me on slack if you need me to do something to assist while you work on this. :-)

PR #718, opened 2015-10-02
Event Slack Notification Bugfix

sampritipanda:
* Move the Twitter Notification code into the controller to remove dependency
* Hangouts without projects will use Hangout Title in the Tweet.
* Refactored the code to move responsibility to TwitterService


spec/services/twitter_service_spec.rb
```diff
@@ -7,6 +7,113 @@
...
+                                            {"id" => "xxx", "displayName" => "Bar Foo",
+                                             "image" => {"url" => ".../s96-c/photo.jpg"},
+                                             "fa" => "false"}, "locale" => "en", "fa" => "false"},
+                                "2" => {"id" => "xxx.id.google.com^xxx", "hasMicrophone" => "true",
+                                        "hasCamera" => "true", "hasAppEnabled" => "true",
+                                        "isBroadcaster" => "true", "isInBroadcast" => "true",
+                                        "displayIndex" => "2", "person" =>
+                                            {"id" => "xxx", "displayName" => "John Doe",
+                                             "image" => {"url" => ".../s96-c/photo.jpg"}, "fa" => "false"},
+                                        "locale" => "en", "fa" => "false"}}
```
yggie: non-blocking issue, but it looks like the participants fixtures are duplicated here, perhaps you could move this into a `let` block a little higher up, since they take up quite a bit of space :grinning:

yggie:
looks good :+1:

PR #741, opened 2016-01-05
Temporarily disable sign up

diraulo:


tansaku:
thanks @diraulo - might it be a good idea to remove the form itself so that real users are not confused?  anyway, we've had at least 10 more sign ups over night, so we need to get something in place ...

diraulo:
@tansaku with this change they'll be redirected to the home page with a flash message saying registration is currently disabled. If it's all good and CI passes i'll merge it and push to production.

tansaku:
right, and github signup will still work @diraulo ?

diraulo:
@tansaku github and g+ signups should still work. I just disabled the email form for now.

PR #748, opened 2016-02-01
[WIP] experimenting with local time support

### LINKED ISSUE #753: Show local time in events/index view
importing from https://www.pivotaltracker.com/story/show/82968236 (requested by amy-lightholder)

As a user, I would like to see the time for listed events in local as well as universal time

Currently, the Show view for events displays the local time but the Index view does not. This is confusing for users, especially during the transition between standard and daylight savings time, which happens on different dates in in different countries. This can be partially solved by displaying the local time in the index view.

dalupus:
work in progress for local time support.  Having issues with repeating events and timezones.

This work closes #753

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -4,7 +4,7 @@ development: &dev
   database: websiteone_development
   pool: 20
   username: postgres
-  password:
+  password: postgres
 
 test: &test
   <<: *dev
--- a/config/database.yml
+++ b/config/database.yml
@@ -4,7 +4,7 @@ development: &dev
   database: websiteone_development
   pool: 20
   username: postgres
-  password: postgres
+  password:
 
 test: &test
   <<: *dev
--- a/app/helpers/event_helper.rb
+++ b/app/helpers/event_helper.rb
@@ -58,4 +58,8 @@ module EventHelper
   def format_date(datetime)
     datetime.strftime('%F')
   end
+
+  def format_local_time(datetime)
+    local_time(datetime,'%l:%M %p (%Z)')
+  end
 end
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -15,7 +15,7 @@
         <% event = instance[:event] %>
         <% time = instance[:time] %>
         <% prefix = time < Time.now ? 'Started at ' : 'Starts at ' %>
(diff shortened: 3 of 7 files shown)


dalupus:
so I am having an issue with repeating events in that it seems to be using the timezone of my system.

So like if we have an event scheduled for 10am UTC weekly, when it gets to the view it is instead 10am eastern time.

I do not see this issue with the one off events.

dalupus:
I have confirmed that it is the timezone of the system which is being used somehow.

tansaku:
thanks @dalupus - really appreciate you getting a WIP pull request up - very helpful.  As I was saying in the standup - I think that my angle of attack here would be to examine the event_specs to as a way in to see how the repeating events are mangling the timezone (which is what I suspect they are doing).

Just repeating the above for the record - will take a more detailed look as soon as I have a spare moment.

tansaku:
Fixes #753

config/database.yml
```diff
@@ -4,7 +4,7 @@ development: &dev
   database: websiteone_development
   pool: 20
   username: postgres
-  password:
+  password: postgres
```
diraulo: @dalupus you usually want to avoid pushing changes to `database.yml`

tansaku:
@dalupus - apologies - I think you did tell us the status of this in a standup - can you remind me where we got to?

Basically there's some problem with the repeated events in ice-cube using the environment timezone in a strange way, that prevents us from using that mechanism to set the timezone of the capybara browser?

So our options are to find another way to set the timezone on the capybara browser: 

http://stackoverflow.com/questions/19532425/how-to-stub-a-specific-timezone-for-capybara-testing

or to find some way to adjust icecube - possibly opening an issue on the icecube repo itself?

https://github.com/seejohnrun/ice_cube/issues/new

I'd say the priority it to get a description of the issue to the ice_cube people ... what do you think?

dalupus:
I would need to create a small example for the ice_cube folks.  This works for events which aren't repeated so I will just add a view test there and move on.  We can add a chore to log a bug if we wish.

config/database.yml
dalupus: yes this was a mistake.  I didn't mean to push that

I am reworking the pull

PR #771, opened 2016-02-06
Fix Blank Youtube Notifications

sampritipanda:
* Add a check in the slack service to make sure that blank notifications don't get sent out to agile-bot

Fixes AgileVentures/agile-bot#2

### DIFF
--- a/app/services/slack_service.rb
+++ b/app/services/slack_service.rb
@@ -3,6 +3,7 @@ module SlackService
 
   def post_hangout_notification(hangout)
     return unless Features.slack.notifications.enabled
+    return if hangout.hangout_url.blank?
 
     uri = URI.parse "#{Slack::BOT_URL}/hubot/hangouts-notify"
     Net::HTTP.post_form uri, {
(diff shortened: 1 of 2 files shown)


sampritipanda:
Build needs to be restarted. The cucumber tests are failing on a completely unrelated test case :disappointed:

diraulo:
thanks @sampritipanda will merge this in ;)

app/services/slack_service.rb
```diff
@@ -3,6 +3,7 @@ module SlackService
 
   def post_hangout_notification(hangout)
     return unless Features.slack.notifications.enabled
+    return if hangout.hangout_url.blank?
```
tansaku: just thinking about this an alternative:

```ruby
return if hangout.blank? :hangout_url
```
which might be combined with ...

```ruby
class EventInstance < ActiveRecord::Base
  def blank?(field)
    field.send(:blank?)
  end
end
```

sampritipanda: Any particular reason why this would be better than the original method?

Also I discovered that `hangout.hangout_url?` == `hangout.hangout_url.blank?` so maybe that could be a alternative as well.

tansaku: the idea is to avoid breaking the law of demeter:

https://en.wikipedia.org/wiki/Law_of_Demeter

relating to not reaching into other objects, but just asking them directly.

If `hangout.hangout_url?` is available that neatly sidesteps the issue and is a double win :-)

PR #772, opened 2016-02-07
82968236 show local time

### LINKED ISSUE #753: Show local time in events/index view

dalupus:
Everything is working for showing local time.   Had to adjust a few tests to adapt to the more normal time display format.

This work closes #753

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -19,7 +19,8 @@ module WebsiteOne
 
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    # config.time_zone = 'Central Time (US & Canada)'
+    #config.time_zone = 'Central Time (US & Canada)'
+    ENV['TZ']='UTC'
 
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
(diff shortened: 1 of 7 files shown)


diraulo:
@dalupus Thanks for this. I will run this locally and merge if all is in order

config/application.rb
```diff
@@ -19,7 +19,8 @@ class Application < Rails::Application
 
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    # config.time_zone = 'Central Time (US & Canada)'
+    #config.time_zone = 'Central Time (US & Canada)'
+    ENV['TZ']='UTC'
```
tansaku: could this be in `config/environments/test.rb` ?  I moved if there and your test (and all the others) still passed.  We only need this for the test environment no?

tansaku:
Overall looks great @dalupus - @diraulo and I both checked - just see minor comment about config above

config/application.rb
dalupus: yes and no.  If this wasn't set the entire site would break if ever run on a non utc server.

Although the only place it gets run on a non-utc server is when testing.

I put it here because it is a system requirement that it be run in the UTC timezone, not just a testing requirement.

dalupus: Actually, yes I definitely want to leave it where it is, because I run this site on a non-utc system.  If not set I couldn't do any type of "manual" testing with events as the times would be all screwed up.

PR #774, opened 2016-02-08
Pushes develop to staging

diraulo:


tochman:
?? What's this?

diraulo:
opened PR to `master` instead of `staging` and inadvertently sync up `master` with `develop` that's why those commits appear here :disappointed:

PR #826, opened 2016-02-21
817 add security auditing to the ci builds

freeranger:
Note this PR adds the tasks but does not put them in the ci build since that would break due to audit failures.
This is because bundler has found a lot of out of date gems with known security vulnerabilities.
Brakeman has found some vunerabilities too but it does not fail the build (or at least not as configured).
You can see sample output here from my PR
https://semaphoreci.com/freeranger/websiteone/branches/817_add_security_auditing_to_the_ci_builds/builds/1

If you don't want to add it to the default rake task but still want to be able to run "on demand" then just remove the final line in both of the new .rake files and you can run 
```
bundle exec rake brakeman:run
```
and
```
bundle exec rake bundler:audit
```
to see the problems


diraulo:
hey @freeranger I see you have commits from other PRs that haven't been merged yet as part of this one ... was that done on purpose?

Since merging this would break the build anyway, wouldn't it be good to fix those issues as part of the PR as well?

freeranger:
Sorry, noob mistake - what's the best way to fix it - kill the PR and start again?
Re: the code issues - no for several reasons
1. This PR is abbut reporting problems with the code base, not fixing them - fixing them is orthogonal to reporting them.  I can take it out of the default rake task so the build won't fail
2. I don't have any experience of the code base or enough experience of Ruby to fix the issues so I wouldn't be comfortable at this stage.

It's mainly point 1 though - the issues are separate and may take considerable time to fix and can possibly be fixed incrementally.

freeranger:
Auditing failures

[bundler.txt](https://github.com/AgileVentures/WebsiteOne/files/139901/bundler.txt)
[brakeman.txt](https://github.com/AgileVentures/WebsiteOne/files/139902/brakeman.txt)

freeranger:
Ok I have reverted the ruby image change stuff accidentally included except for the removal of   app/assets/images/AV205X100.png which I can't seem to fix - locally the file exists but I am up to date with the remote origin so I don't know what's going on - it looks like the PR is still deleting it??

I have taken the checks out of the default rake task - someone with the right privilege would have to add them into the semaphore build but perhaps you don't want to do that anyway until the issues are fixed, which I would still argue is a completely separate PR

freeranger:
Also, I don't see how this is breaking the build now - there are no actual code changes, no new tests, nothing that should be running.  Either the failure is unrelated or somehow that image file I can't restore is doing it?

freeranger:
Hmm reran the same build on my semaphore instance and it just worked...seems to be a transient issue...

freeranger:
Something went a bit wrong - I will resubmit - can't seem to recover image I deleted as part of accidental inclusion of other PR :/

PR #853, opened 2016-03-02
852 Fix Pivotal tracker url validator bug

### LINKED ISSUE #852: Pivotal tracker url validator bug
imported from https://www.pivotaltracker.com/story/show/90656646

diraulo:
fixes #852


tansaku:
if we were being fussy I guess we could change the branch name here to 759_fix_pivotal_link_bug or 759_add_support_for_new_pivotal_links, but probably not worth it in this instance.

Something we could also have discussed in the retro/kick-off is whether we should delete branches more aggressively ...

features/projects.feature
```diff
@@ -118,6 +118,29 @@ Feature: Create and maintain projects
     And I should see a link to "Title New" on github
     And I should see a link to "Title New" on Pivotal Tracker
 
+  Scenario: Saving a new project with new PivotalTracker link format: success
```
tansaku: we've completely replicated a scenario here with only a single minor difference  It might be DRYer to use a scenario outline here:

https://github.com/cucumber/cucumber/wiki/Scenario-Outlines

spec/models/project_spec.rb
```diff
@@ -33,7 +33,7 @@
       expect(subject).to_not be_valid
     end
 
-    it "should not accept invalid Pivotal Tracker URL" do
+    it 'should not accept invalid Pivotal Tracker URL' do
```
tansaku: is the s format now completely invalid? looks like it redirects to the n format - I just wonder if some old projects that still have the s might encounter problems - easy fix I guess for the maintainer I guess

diraulo: the `s` format is not invalid, Pivotal tracker will automatically redirect those links to the `n` format

tansaku: right, but this test expects the s version to not be valid ...

features/projects.feature
diraulo: @tansaku will refactor this and push an update ;)

spec/models/project_spec.rb
diraulo: @tansaku note that the link tested here has an extra level before the `project_id` which makes it invalid. A valid link should be of this format
```
https://www.pivotaltracker.com/s/../912312
```

and not

```
https://www.pivotaltracker.com/s/../../912312
````
It's not really about it having the s.

diraulo:
Closing this in favor of #860

PR #856, opened 2016-03-03
WIP: 855 convert should to expect

### LINKED ISSUE #855: convert legacy ".should" assertions to "expect.to" syntax
pulled out of #749 because this will be a relatively large commit

freeranger:
Closes #855


freeranger:
Hmm that went a bit wrong - I have commits from the original branch - oops

PR #848, opened 2016-02-29
749 rspec junk

freeranger:
Closes #749

PR #858, opened 2016-03-03
855 convert should to expect

### LINKED ISSUE #855: convert legacy ".should" assertions to "expect.to" syntax

freeranger:
Closes #855

Convert should and should_receive to new expect syntax to remove deprecation warnings from the rspec log.
Note: this is all test changes - no application code has been touched

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,4 @@
-2016-02-11 @freeranger - Updated the RubyMine logo
-2016-03-03 @freeranger - Remove junk from rspec log
-2016-03-13 @freeranger - Fix Sponsor alignment
+2016-02-11 @freeranger #790 - Updated the RubyMine logo
+2016-03-03 @freeranger #749 - Remove junk from rspec log
+2016-03-13 @freeranger #855 - Convert 'should' to 'expect'
+2016-03-13 @freeranger #900 - Fix Sponsor alignment
--- a/features/step_definitions/articles_steps.rb
+++ b/features/step_definitions/articles_steps.rb
@@ -24,7 +24,7 @@ Then(/^I should see a preview containing:$/) do |table|
   preview_window = windows.last
   page.within_window preview_window do
     content.each do |text|
-      page.should have_text text
+      expect(page).to have_text text
     end
   end
 end
--- a/features/step_definitions/avatar_steps.rb
+++ b/features/step_definitions/avatar_steps.rb
@@ -1,5 +1,5 @@
 Then /^I should see my avatar image$/ do
-  page.should have_xpath("//img[contains(@src, '#{@avatar_link}')]")
+  expect(page).to have_xpath("//img[contains(@src, '#{@avatar_link}')]")
 end
 
 Then /^I should see "([^"]*)" avatars$/ do | arg |
--- a/features/step_definitions/basic_steps.rb
+++ b/features/step_definitions/basic_steps.rb
@@ -242,9 +242,9 @@ Then /^the "([^"]*)" field(?: within (.*))? should( not)? contain "([^"]*)"$/ do
     field_value = (field.tag_name == 'textarea') ? field.text : field.value
     field_value ||= ''
     unless negative
-      field_value.should =~ /#{value}/
+      expect(field_value).to match(/#{value}/)
     else
-      field_value.should_not =~ /#{value}/
+      expect(field_value).to_not match(/#{value}/)
     end
   end
 end
@@ -279,7 +279,7 @@ end
 
 When(/^I should see a selector with options$/) do |table|
   table.rows.flatten.each do |option|
-    page.should have_select(:options => [option])
+    expect(page).to have_select(:options => [option])
   end
 end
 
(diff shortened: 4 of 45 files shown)


diraulo:
Thanks for this @freeranger. After we merge this, shouldn't we enforce the `expect` syntax? I recall in #848 you're allowing the use of the `should` syntax to get rid of the warning messages.

freeranger:
IIRC that didn't actually suppress all the messages that this PR fixes.

Perhaps get both merged in and then see if we can remove the use of should?

tansaku:
so this branch now needs to be synced to upstream.  It's an opportunity to try `git rebase -i upstream/develop` ... if you fancy a challenge :-)

freeranger:
@tansaku it's just not working for me - I do a rebase and it complains about conflicts, I fix them, continue the rebase, push, rebase again and the same conflicts reappear.  Now the history of this PR looks a right mess
I could just create a brand new PR unless you want to have a go at doing the rebase yourself and tutor me in what's going wrong?  Or shall I just go ahead and create a new PR from a clean checkout?

freeranger:
Actually I think it's good to go now - I was doing the rebase and then git was telling me my local repo was behind the remote one...so i did a pull and was back to square one.
What I did, rightly or wrongly, is
* `git rebase -i upstream/develop`
* `git push -f` to force it into the remote branch

This seems to have worked.  I have lost all my history now, but hey ho, the price of a clean PR :}
I was happy to leave this as a two commits, which could go separately but my fannying about earlier trying to do the rebase left it all in a bit of a mess so I collapsed it down into one.
Anyway, please review and hopefully accept the PR finally!

tansaku:
great work getting the rebase done @freeranger - and great work on this PR.

There's a really interesting issue here regarding the tension between wanting to keep the commits and history in the PR and wanting to get a clean commit history and project.

I'm not sure if it's actually that helpful to the learner, but I imagine that PRs maintaining their history and individual commits are a better learning resource (although how many learners will actually make use of them), than rebased ones; and in principle we could rebase in a fresh branch and PR, but that is possibly even more convoluted ...

Maybe learning the hang of rebasing is just pain everyone has to go through, and we should prioritize a clean commit history for the project (supporting bisects, better debugging) and be deleting old branches rigorously to try and reduce the overall complexity of the repo ...

tansaku:
@diraulo code and rspec log for this are looking good - want to double check for me and I think we can get this merged?

freeranger:
@tansaku Thanks - is that the correct way then, to do a `push -f` after the rebase?

If we always do a 1 commit PR then doesn't it make it harder to do a bisect because the commit that caused the problem is the whole thing, whereas in reality it was the 5th out of 12 commits or whatever...
I can understand keeping things like "added forgotten file" or "fixed broken test" out of the logs because these things shouldn't happen, but where an issue is resolved in incremental, releasable steps, it seems to me there can be value in keeping that history.

Perhaps it's one of those "decide on an individual basis" things rather than a rule across the board.

But yes, people will need to get the hang of rebasing....it can just be a bit scary, especially when you're not a command line jockey :}  I ditched all those skills years ago except somehow I can still remember `<esc>:wq` for vi and friends :D

tansaku:
well leaving aside what is "correct", if you do a `push -f` to the feature branch after the rebase then the previous commits will be swept away, and the current PR will now consist of only the commits left after the rebase.  This is what quite a few open source projects request in terms of submissions, but not all.

A 1 commit PR does in principle make a bisect harder.  The key thing for bisect is that all commits should be green.  As you probably saw as you did an interactive rebase, it's up to you whether you go from 10 commits to 1, or from 10 to 7 or 10 to 3 - I think that's a judgement call.

Personally I'd be happy merging a PR with 3 commits as long as all three of them are green and with sensible commit messages, although 3 commits begs the question about whether they should have been 3 pull requests ...

The bigger problem for us is that our tests fail intermittently making it difficult to judge.  I suspect the capybara upgrades in your other PR will help, and I am also starting to wonder if it's puffing_billy sandboxing that might close the gap for us ... see #904

freeranger:
If I don't do a `push -f` then how else would I get the rebased code to the PR?
I would be happy for my branch to keep the full history and for the PR to have a slimmed down one, but is that even possible?

re: commits vs separate PRs - it depends.  the issue is "clean up the logs" and one commit will clean up certain things, another will clean up other things - so they are both valid in their own right in the context of the issue and you could cherry pick one commit over another if required....but is there any real value in separating them out into completely separate issues?
Again probably a case-by-case thing rather than having a blanket rule...

tansaku:
you have to do a `push -f` to get a completely rebased code to the PR.  The alternative is `checkout -b same_number_related_story_description` and push the rebased code to that and open a new PR; however that feels like a fair amount of hassle for an unclear amount of gain.

And yes, I think we need to be flexible and deal with things case by case ...

diraulo:
@freeranger I just merged one of your PR and this one now has some conflicts. Could you please look into that so I can merge this one too.

tansaku:
@diraulo I think freeranger has limited access to internet today - @freeranger if you give me collaborator access on your WSO fork I can sort this out for you (do I have that access already perhaps?)

tansaku:
@diraulo @freeranger actually I just did a separate PR #905

sorry @freeranger if I'm stealing your thunder there - just keen to reduce our PR backlog and move on to other things - if we are happy with everything we should be able to close this and merge #905

diraulo:
@tansaku I think the conflict is coming from the `changelog.md`

tansaku:
@diraulo it was more than that - there were conflicts in the cucumber steps that I fixed in #905 ...

tansaku:
and I resolved them, but not correctly in #905 - thought it was a quick fix, but no - let's leave this for freeranger to sort out when he has a moment

freeranger:
@diraulo @tansaku Ok, back to green - all checks passed, and NO conflicts - if one of you guys would like to pull this in before anything else goes wrong, that would be great ;)

PR #864, opened 2016-03-04
Consolidates getting started documentation

### LINKED ISSUE #863: Consolidate "getting started" and "project setup" documentation
There are various getting started and setup docs in the Github wiki and the AV site wiki, let's consolidate them into a single coherent set of docs in the main GitHub repo where we can have more active review to ensure that they don't go out of sync.

tansaku:
fixes #863

When this gets merged we should ensure that the relevant pages get deleted in the:

* [ ] Github Wiki
* [ ] AV Wiki

### DIFF
--- a/docs/osx/el_capitan_10.11.2_install_notes.md
+++ b/docs/osx/el_capitan_10.11.2_install_notes.md
@@ -1,6 +1,8 @@
 WebSiteOne install
 ————————
 
+:warning: These are notes from @tansaku. Use [the project setup guide](docs/project_setup.md) to setup your own environment. Refer to these notes if you encounter issues mentioned below where solutions are not covered in the project setup guide. :warning:
+
 Fresh 10.11.2 (El Capitan) OSX
 
 1) Using git for the first time prompted installing developer tools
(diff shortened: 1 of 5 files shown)


tansaku:
@diraulo and anyone else interested - I'm breaking the documentation issue into sub-issues so they can be addressed with separate pull requests.  I think this PR could be merged now?  There's probably room for further consolidation, but I think I've removed the specific inaccuracies ...

docs/osx/el_capitan_10.11.2_install_notes.md
```diff
@@ -0,0 +1,87 @@
...
+————————
+
+Fresh 10.11.2 (El Capitan) OSX
+
+1) Using git for the first time prompted installing developer tools
+
+2) pull down code
+
+```
+git clone http://github.com/AgileVentures/WebsiteOne
```
diraulo: @tansaku aren't we supposed to clone from our own fork?

tansaku: yes, - although those are my rough notes from what I did personally which was work off the main repo, which when I'm project manager I always do.  

if these are polished up to be better notes then we should probably change that - but I wonder if we should even include this at all?  maybe I should put a :construction: at the top or something ...

diraulo: @tansaku I added a note at the top of this document. If that's fine we can merge it. As we progress with the documentation, we should have a doc that covers common issues during setup. Then we extract the ones you have here with solutions into that document.

PR #879, opened 2016-03-08
Pushes develop to staging

diraulo:


diraulo:
@tansaku Should we then merge this since #880 was already an existing bug. I held off on this earlier on to check that out first.

tansaku:
@diraulo not much point I think as it won't change anything on the server - well I guess it's integration testing - we'll look at it in more detail tomorrow ...

tansaku:
oh right sorry, yeah the rest can go up ...

PR #894, opened 2016-03-11
Appropriate defaults for creating events

### LINKED ISSUE #891: setting appropriate defaults (specifically default end for repeats) for event creation
We need appropriate defaults for event creation to avoid repeats hanging around indefinitely

repeat defaults to never ending

Currently as shown above, selecting a repeat leads to a never ending event ...

mtc2013:
fixes #891

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -9,6 +9,7 @@ class Event < ActiveRecord::Base
   validates :name, :time_zone, :repeats, :category, :start_datetime, :duration, presence: true
   validates :url, uri: true, :allow_blank => true
   validates :repeats_every_n_weeks, :presence => true, :if => lambda { |e| e.repeats == 'weekly' }
+  validates :repeat_ends_on, :presence => true, :allow_blank => false, :if => lambda{ |e| e.repeats == 'weekly' and e.repeat_ends_string == 'on'}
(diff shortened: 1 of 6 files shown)


mtc2013:
@tansaku @diraulo This is ready for review.

diraulo:
screenshot 2016-03-15 14 01 53

@tansaku @mtc2013 Why show `End Time in UTC` field if the event is set to never repeat? Also wouldn't it be preferable to have it say `End date`?

diraulo:
@tansaku have you seen my previous comment? I see you've solved the issue with confusing field in #907

tansaku:
@diraulo yeah - it has been like that for a long time - that was another long standing UI issue that as you noticed we addressed in #907, so perhaps this can be merged?

diraulo:
@tansaku what about displaying the field when the event is set to never repeat? Shouldn't it be hidden in that case?

tansaku:
@diraulo - yes, but that's what the other PR fixes.

I believe problems you are highlighting here were already problems before we started working on fixing the defaults.

I think this PR in combination with the other fix all the concerns - unless I'm missing something ...

diraulo:
Okay then I'll review the other PRs and merge this one too

PR #907, opened 2016-03-15
Rename End Time in UTC to End Date and hide appropriately

### LINKED ISSUE #893: "End Time in UTC" field is confusing
"End Time in UTC" field is confusing

This should be "End Date" and should only be visible when an event is set to repeat

tansaku:
fixes #893


diraulo:
@tansaku I have some merge conflicts here after merging #894

tansaku:
@diraulo fixed

PR #913, opened 2016-03-16
744 - Displaying long nicknames

### LINKED ISSUE #744: User's index visual improvements for long nicknames
As you can see in this screenshot:
 now_chrome 

long nicknames are breaking the layout in large screen.
You can easily fix this setting: `.media-body { 
display: block;
max-width: 100%;
} `
Here is the result on chrome:
after_chrome

You can also visualize the whole nickname changing the .index-summary class:
`.index-summary { overflow: auto; word-wrap: break-word }`
and here is the result:
last_chrome

I would make a pull request but i don't really know which .css i should change :sweat_smile:

freeranger:
closes #744

### DIFF
--- a/app/assets/javascripts/users.js
+++ b/app/assets/javascripts/users.js
@@ -12,6 +12,12 @@ WebsiteOne.define('Users', function () {
         }, 300);
     }
 
+    function enableBootstrapTooltipsOnNonTouchDevices() {
+        if (!('ontouchstart' in document.documentElement)) {
+            $('[data-toggle="tooltip"]').tooltip({delay: { "show": 500, "hide": 100 }});
+        }
+    }
+
     function init() {
         $('.yt_link').on('click', selectVideo);
 
@@ -65,6 +71,8 @@ WebsiteOne.define('Users', function () {
         });
 
         $('#require_users_profile').modal({show: true});
+
+        enableBootstrapTooltipsOnNonTouchDevices();
     }
 
     return {
--- a/app/views/users/_user_list.html.erb
+++ b/app/views/users/_user_list.html.erb
@@ -6,7 +6,7 @@
                 <%= presenter.gravatar_image(size: 80,
                                              class: 'img-rounded media-object pull-left') %>
                 <div class="media-body index-summary">
-                  <p class="media-heading"><%= presenter.display_name %>
+                  <p class="media-heading" data-toggle="tooltip" title="<%= presenter.display_name %>"><%= presenter.display_name %>
                     <% if user.online? %>
                         <%= image_tag('green-dot.png', size: '10x10', alt: 'Online!') %>
                     <% end %></p>
@@ -14,7 +14,7 @@
                       <p><%= presenter.title_list %></p>
                   <% end %>
                   <% if user.country_name %>
-                      <p><%= presenter.country_name %></p>
+                      <p data-toggle="tooltip" title="<%= presenter.country_name %>"><%= presenter.country_name %></p>
                   <% end %>
                 </div>
               </a>
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -437,7 +437,7 @@ When(/^I select "(.*?)" from the "(.*?)" list$/) do |selected_from_list, list_na
   page.select(selected_from_list, from: filter)
 end
 
-Given(/^I have an incoplete profile$/) do
+Given(/^I have an incomplete profile$/) do
   @user.bio = ''
   @user.save
 end
--- a/app/assets/stylesheets/global/users.scss
+++ b/app/assets/stylesheets/global/users.scss
@@ -104,6 +104,11 @@
   }
 }
 
+.media-body {
+  display: block;
+  width: 60%;
+}
+
 .media-item {
   height: 100px;
   margin-bottom: 30px;
--- a/changelog.md
+++ b/changelog.md
@@ -3,3 +3,4 @@
 2016-03-13 @freeranger #855 - Convert 'should' to 'expect'
 2016-03-13 @freeranger #900 - Fix Sponsor alignment
 2016-03-15 @freeranger #849 - Too much whitespace in title on mobile
+2016-03-17 @freeranger #744 - Truncate long names and countries
--- a/features/remind_complete_profile.feature
+++ b/features/remind_complete_profile.feature
@@ -6,7 +6,7 @@ Feature: Remind User complete profile
 
   Scenario: User has incomplete profile from home page
     Given I exist as a user
-    And I have an incoplete profile
+    And I have an incomplete profile
     When I sign in with valid credentials
     Then I should see "We've noticed that your profile is incomplete."
     And I should see link "Take me to my profile page"
@@ -14,14 +14,14 @@ Feature: Remind User complete profile
 
   Scenario: User has incomplete profile from home page and choose 'edit registration'
     Given I exist as a user
-    And I have an incoplete profile
+    And I have an incomplete profile
     When I sign in with valid credentials
     And I click "Take me to my profile page"
     Then I should be on the "edit registration" page
 
   Scenario: User has incomplete profile from home page and choose 'continue'
     Given I exist as a user
-    And I have an incoplete profile
+    And I have an incomplete profile
     When I sign in with valid credentials
     And I click "Continue using the site"
     Then I should be on the "home" page
@@ -33,7 +33,7 @@ Feature: Remind User complete profile
 
   Scenario: User has incomplete profile not from home page
     Given I exist as a user
-    And I have an incoplete profile
+    And I have an incomplete profile
     And I visit "projects"
     When I sign in with valid credentials
     Then I should see "We've noticed that your profile is incomplete."


tansaku:
@freeranger you have some unimplemented step definitions causing this build fail

freeranger:
@tansaku I do indeed - it's a "Work in Progress" :)
This isn't one of the failing builds I was talking about earlier.  I knew this would break...

freeranger:
@tansaku @diraulo This is ready to go I think.
There doesn't seem to have been an automatic heroku deployment for some reason?
I did one myself anyway to: http://freeranger-wso-longnames.herokuapp.com
Only problem is there are no users and no way to sign up since email registrations are disabled and I guess I don't have the relevant github/G+ keys to allow registrations to happen from this instance..

I did country names as well as user names, since they were a problem also, and I have tooltips with the full text only where a truncation occurred.

You can see examples of name and country truncation in this screenshot:
image

tochman:
You can seed your db and create some users. You know that, right?

freeranger:
Yes I knew about the seeds file but I didn't know I could ssh into the heroku instance and run it.
Have done now, live site is at http://freeranger-wso-longnames.herokuapp.com/users
@diraulo @tansaku - this PR is good to go if you like the look of it.  The semaphore failure for the checkin is one of the usual intermittent failures...

tansaku:
yeah, maybe we could add `rake db:seed` to the app.json file @diraulo ?

tochman:
You don't have to ssh into heroku. Use the cli
$ heroku run rake db:seed

tochman:
@tansaku wouldn't that seed in production as well?

tansaku:
ooh, maybe not sure - I think app.json only runs for the automated builds on PRs but we would need to check carefully ...

diraulo:
Yesterday I added `rake db:setup` which will run migration and seed the db as well

diraulo:
@tansaku @tochman the `app.json` file is only for automated deployment of PRs on Heroku

For some reason this one didn't kick-off .. perhaps @freeranger didn't have the file on his repo when he started working on this

freeranger:
@diraulo That is possible, though I did do a rebase to develop when I checked in so would have picked it up then - perhaps the file needs to exist when the PR is first created?

diraulo:
Thanks for this @freeranger fixes the issue :smile: 

Just one concern before I merge this in. Why did you choose this approach and not just the CSS fix that was suggested in the issue?

freeranger:
Hi @diraulo, several reasons really:
- It's not 100% cross browser, though granted this is becoming less of an issue
- The actual text in the page source is still the full text so I don't know how I would test that it shows the ellipses
- I only wanted to add a title if the ellipses are used and I can't guarantee the css ellipses and ruby ellipses usage would coincide at the same string length

diraulo:
Fair enough @freeranger - @tansaku any objections to merging this one?

tansaku:
@diraulo can you leave it with me for a more detailed review?

tansaku:
@freeranger - I've given this a closer look and I'm perfectly happy with the non-CSS approach for consistency.  However see my inline comments

app/views/users/_user_list.html.erb
```diff
@@ -6,7 +6,7 @@
                 <%= presenter.gravatar_image(size: 80,
                                              class: 'img-rounded media-object pull-left') %>
                 <div class="media-body index-summary">
-                  <p class="media-heading"><%= presenter.display_name %>
+                  <p class="media-heading" title="<%= presenter.display_name unless presenter.display_name.length < 15 %>"><%= truncate(presenter.display_name, length: 14) %>
```
tansaku: this feels like too much logic in the view - either pull to a helper, or check if you even need the `unless` statement

app/views/users/_user_list.html.erb
```diff
@@ -14,7 +14,7 @@
                       <p><%= presenter.title_list %></p>
                   <% end %>
                   <% if user.country_name %>
-                      <p><%= presenter.country_name %></p>
+                      <p title="<%= presenter.country_name unless presenter.country_name.length < 15 %>"><%= truncate(presenter.country_name, length:14) %></p>
                   <% end %>
```
tansaku: this feels like too much logic in the view - either pull to a helper, or check if you even need the unless statement

app/views/users/_user_list.html.erb
```diff
@@ -14,7 +14,7 @@
                       <p><%= presenter.title_list %></p>
                   <% end %>
                   <% if user.country_name %>
-                      <p><%= presenter.country_name %></p>
+                      <p title="<%= presenter.country_name unless presenter.country_name.length < 15 %>"><%= truncate(presenter.country_name, length:14) %></p>
```
tansaku: this feels like too much logic in the view - either pull to a helper, or check if you even need the unless statement

features/step_definitions/user_steps.rb
```diff
@@ -441,3 +441,15 @@
   @user.bio = ''
   @user.save
 end
+
+When(/^I look at user "(.*)"/) do |email|
+  @user = User.find_by_email email
+end
+
+Then(/^the tooltip (?:.*) (should not|should) be "(.*)"/) do |option, text|
```
tansaku: rather than checking the entire page, shouldn't we try to check that the tooltip occurs in the right place using `within`?

app/views/users/_user_list.html.erb
freeranger: @tansaku it's marginal but I take your point....how much logic is too much logic...a debate for another day ;)

There are several options here:
- I can always display the title, which gives more consistency to the experience (at the expense of a larger download)
- I can move the title thing to a helper, but then the actual display name truncation still happens separately so...another helper?  
- I could have a single helper for the entire <p class=.....?>, including the html

Also by "helper" do you mean in a separate helper file or on the User class itself?  If the latter then something like presenter.short_display_name?

features/step_definitions/user_steps.rb
freeranger: @tansaku - Could do yes - though for the negative case shouldn't the check still be the whole page?  I want to make sure that tooltip does not exist anywhere, whether it is the right place or not.

What would you limit it to?  `within ('#UsersList')`?  would that work or do I need to search to find #UsersList and then search within that?  
Also since it looks like I have to revisit anyway, I'd like to introduce https://github.com/kucaahbe/rspec-html-matchers and use have_tag (assuming it will work) since semantically that's what I'm looking for - a tag with a specific title.  I don't like the xpath expression I used because it doesn't express the intent well enough IMO

freeranger:
semaphore build failed due to intermittent failure `Scenario: Closing an existing video and opening a new one should update the player`
A rebuild would probably fix that...

app/views/users/_user_list.html.erb
tansaku: ah, follow up on this as I look closer - you are setting tooltips only when they are long - why not reduce the complexity and set them all the time, then we don't have to worry about testing for both cases - just have them always there?

freeranger: We could - I was worried about making that page bigger and therefore slower than it already is.
hmmm...on mobile it's even more of an issue - you just won't want the tooltips at all because most mobiles won't support the idea of "hover over" in order to see the tooltip....though mobile is a separate PR :}
So how about I always add the tooltip on md and lg screen sizes but never add it on sm and xs?

freeranger:
@tansaku @diraulo - Ok hopefully the final changes to this (barring a squash if this is good enough to go).
The tooltip will always be added regardless of name/country length.
It will use bootstrap tooltip styling, though only for non touch devices as on touch (typically phone or tablet) it apparently causes problems when you come to click on it.

I changed it to be css based in the end because it was looking really naff on mobile if you cut off at 15 characters but have a load of space on the device (because it is now only showing one user per line).
I had to remove my tests because I can't test that the browser has applied the ellipses styling due to the size of the window, but I think the benefits of a better UI experience outweigh the downsides of not having a test for it.

I also (and I realise this is scope creep) fixed a typo in a step definition.

If you are happy with this implementation (see demo at http://freeranger-wso-longnames.herokuapp.com/users) then I will squash the commits and you can merge it in.

app/assets/javascripts/users.js
```diff
@@ -65,6 +65,11 @@ WebsiteOne.define('Users', function () {
         });
 
         $('#require_users_profile').modal({show: true});
+
+        // Enable bootstrap tooltips on non-touch devices only
+        if (!('ontouchstart' in document.documentElement)) {
+            $('[data-toggle="tooltip"]').tooltip({delay: { "show": 500, "hide": 100 }});
```
tansaku: this looks like it is just adjusting the tooltip delay rather doing the enabling mentioned in the comment ...

freeranger: @tansaku No, it enables the tooltip and passes in an option to adjust the delay as part of the enabling: http://getbootstrap.com/javascript/#tooltips-options
You will see the tooltip working if you checkout  http://freeranger-wso-longnames.herokuapp.com/users

freeranger:
I have squashed this now so it's ready to go if you are happy with the implementation

app/assets/javascripts/users.js
tansaku: yes, I also see the tooltip working if I comment out that javascript, so I don't think it is enabling the tooltip:

https://developer.mozilla.org/en/docs/Defining_Cross-Browser_Tooltips

I'm also a little concerned about a comment that describes what a piece of code is supposed to do, and I'm not yet convinced that the code that follows does what is described in the comment.

Either way we should avoid comments and prefer to write method names that accurately describe what code does, e.g. 

```javascript
function addToolTipDelayToNonTouchDevices(){
  if (!('ontouchstart' in document.documentElement)) {
    $('[data-toggle="tooltip"]').tooltip({delay: { "show": 500, "hide": 100 }});
  }
}
```

This reduces the chance of the comment going out of date, and also provides us with methods that can be tested independently

freeranger: @tansaku - You might see tooltips but do you see bootstrap tooltips?  That is the primary function of this bit of code, to enable bootstrap tooltips - it passes in options to control the delay between showing and hiding them, but it is necessary to call this function according to the bootstrap documentation: http://getbootstrap.com/javascript/#four-directions - look at the "opt in functionality" bit.
The comment was more about the why it is doing this than what it is doing but ok, I could refactor to a function called
`initialiseBootstrapsTooltipsForNonTouchDevices`, or would you prefer `initialiseBootstrapTooltipsForNonTouchDevicesWithDelayBeforeShowing`?
The touch related js is all over the shop - stack overflow, blogs, github... https://www.google.co.uk/search?q=%27ontouchstart%27+in+document.documentElement

tansaku: understood, but do we need bootstrap enabled tooltips rather than just plain tooltips?

freeranger: Isn't that question a bit like asking "do we need to use bootstrap &lt;insert name of any bootstrap feature we currently use here&gt;"?   
Do we _need_ to use them?  No.  We _choose_ to use features of bootstrap because we like how they look or they make it quicker/easier to do something - so we use the grid layout, the nav bars and other things bootstrap provides.

I used bootstrap tooltips because they look a lot better (IMO) than the default ones but I can certainly remove them if you prefer the standard look - though I would say if you want WSO to showcase what we can do, then default tooltips is not going to wow anyone :)

freeranger:
Moved tooltip enabling to a separate function

tansaku:
interesting - through squashing the commits we lose the discussion on the commits ...

@diraulo I still have a slight reservations about introducing javascript without unit test, although I guess these are covered by @javascript enabled acceptance tests; but if you're happy to pull this in let's get it merged.

freeranger:
@tansaku I'd be happy to add a test if you can suggest a way to test this?

tansaku:
well there's jasmine for JS unit testing (although I have mixed feelings about that), and now we have a method then it could have a jasmine test with appropriate stubbing, but I'd say that that should be for another pull request maybe?

At the risk of re-igniting our earlier debate I know you prefer the bootstrap styling, however I don't feel strongly about it either way at the moment.  I'd love the design of the site to impress people but there are much bigger UX fish and technical debt issues that are a bigger concern at the moment. 

If it was me submitting this I might have left that bootstrap/JS off for a second PR, i.e. get the default tooltips in there - get that merged and then we could do a deeper dive with javascript/bootstrap and JS testing in another PR without holding up the basic fix.

We want to be careful that in our desire for the site to be awesome that we don't push the technical debt even higher :-)  We're already sitting on a keg-load ...

For simplicities sake I'm happy to merge this as is and look at JS testing in another PR.

Your hard work on all this is very much appreciated of course.

freeranger:
I am somewhat familiar with jasmine, it was more about how specifically to test this particular thing - ie that tooltips only appear on non touch devices, which is the outcome we want to test.

I am not particularly wedded to bootstrap - it serves a purpose - but if we are using it then why refrain from using it is what I was getting at - it argument against using it (and customising it) didn't hold up in light of the fact that we were already using it and customising it.

If you're happy to merge this then that's great but I am still interested in how I would test this - AV is afterall partially about devs learning/honing their craft as well as producing actual useful and usable sites! 

Assuming I could fake the presence of the ontouchstart element, what could I actually usefully test?
I could...delve into the innards of the tooltip method and figure out what it actually does and then do some kind of test for that, but that strikes me as incredibly brittle and I shouldn't really be testing the innards of a third party control anyway....

tansaku:
> I am somewhat familiar with jasmine, it was more about how specifically to test this particular thing - ie that tooltips only appear on non touch devices, which is the outcome we want to test.

> I am not particularly wedded to bootstrap - it serves a purpose - but if we are using it then why refrain from using it is what I was getting at - the argument against using it (and customising it) didn't hold up in light of the fact that we were already using it and customising it. But as you say lets not reignite the debate."

It seems to me that making the counter-assertion here you are reigniting the debate :-) And I find that I cannot help but try to correct the faulty assertion I think you are making.  My point is that this is not about whether to use bootstrap or not - clearly we are using it (and customising it) - but whether we are muddying the water in *this* PR by adding it.  It's clearly not required to complete work on the original ticket.  Every single little extra thing is one more bit of complexity to hang ourselves on.  The entire project teeters on the brink of being manageable because in the past it has leant towards the more liberal, "yeah let’s throw it in" approach.  Given that history we want to be extremely careful about what we add.  

> If you're happy to merge this then that's great but I am still interested in how I would test this - AV is after all partially about devs learning/honing their craft as well as producing actual useful and usable sites!

> Assuming I could fake the presence of the ontouchstart element, what could I actually usefully test?
I could...delve into the innards of the tooltip method and figure out what it actually does and then do some kind of test for that, but that strikes me as incredibly brittle and I shouldn't really be testing the innards of a third party control anyway....

Of course not - a sensible unit test here would stub the 3rd party elements and check that the logic we are adding is operating correctly.  Note the all the following code segments are incomplete sketches trying to give you a general sense.

You might try like this

```javascript

    function enableBootstrapTooltipsOnNonTouchDevices() {
        if (!('ontouchstart' in document.documentElement)) {
            $('[data-toggle="tooltip"]').tooltip({delay: { "show": 500, "hide": 100 }});
        }
    }
```

```javascript

  spyOn(document, 'documentElement').andReturn('ontouchstart');
  spyOn($, 'documentElement').andReturn('ontouchstart');

```

but you would fall foul of the difficulty of stubbing the DOM, and you might reflect on the maxim that you should only stub what you own: https://blog.8thlight.com/eric-smith/2011/10/27/thats-not-yours.html, and might lead you to refactor like so:


```javascript

    function isTouchDevice() {
      return 'ontouchstart' in document.documentElement;
    }

    function tooltipElement() {
      return $('[data-toggle="tooltip"]');
    }

    function enableBootstrapTooltipsOnNonTouchDevices() {
        if (!isTouchDevice()) {
            tooltipElement().tooltip({delay: { "show": 500, "hide": 100 }});
        }
    }
```


```javascript

  var element = {};
  spyOn(Users, 'isTouchDevice').andReturn(false);
  spyOn(Users, 'tooltipElement').andReturn(element);
  spyOn(element, 'tooltip')

  ...

  expect(element).toHaveBeenCalledWith({delay: { "show": 500, "hide": 100 }});

```

and now you could create tests to check the operation of both of the boolean outcomes of your test, only one of which may be being tested in the current acceptance tests.  And this is an interesting example of tests driving us towards a cleaner code design.

There are other options such as trying to get Capybara to emulate the touch setup, but those are of course acceptance tests rather than unit tests.  There's a bigger debate to be had about the different approaches here, but I'd really rather have them in a different PR that is focused on them, and quickly pull in the smallest and least controversial aspect of the original PR.

So, bottom line, by introducing this little aspect of bootstrap you've making the solution to the original underlying problem more complex that it needs to be for an immediate fix, and that is dangerous; in particular in terms of delaying the pulling in of the code.  I was just saying the following to a contributor in LocalSupport about the priorities as regards pull requests:

1) ensure the Pull Request is the smallest chunk possible (splitting tickets if necessary)
2) ensure the user experience quality is high
3) ensure the code quality is high
4) ensure the Pull Request corresponds to the letter of the ticket

All four of the above are very important, but I definitely put the very highest premium on 1) over 2), 3) and 4) ... The longer a pull request hangs around with extra chat going on, the greater the danger of merge conflicts and it ultimately not being merged in.

I also note that we are explicitly trying to move WSO away from being a "entry-level learning"  project because of the overhead associated with negotiating with developers to ensure they are submitting simple code that does not increase the projects' technical debt.  Now I know you came to WSO at my suggestion and may not even have picked up on our intention that WSO should no longer be considered an "entry-level" project, and so my apologies (and the efforts I'm going to to try and explain my philosophy).

My preference was for you to remove this bootstrap javascript to another pull request, where we could start exploring the jasmine tests that I am talking about above.  I would have preferred to have this debate there, or possibly even in an explicitly "entry-level" project like LocalSupport, where I think we have better JS testing stability. I don't want to stifle debate, but I think we need to be cautious about the costs associated with debate.  Certain levels of debate can be exhausting to those participating, and may serve as a disincentive to team members who might then delay or simply give up on trying to offer helpful suggestions.

An important lesson I've learned is that if one strongly disputes people's criticisms or suggestions then our colleagues may well end up being disincentivized to give criticism or suggestions in future. Sometimes it's more efficient to just try what they are suggesting, or spend extra time reflecting on the criticism before responding, rather than quickly defending oneself or justifying ones actions ... at least that's a breakthrough learning I've had in terms of my human interaction recently.

freeranger:
Ok so does this: "For simplicities sake I'm happy to merge this as is and look at JS testing in another PR." still apply or do you want me to pull it out of this PR now? I'm confused now.
I think the plan to move away from WSO being an "entry level" project came into force after I started working on WSO but I'm happy to step away from "WSO" if you prefer.
I don't know what "entry level" means in this context though because I am not new to development by any stretch of the imagination or new to web development either.  New to ruby? yes.  To capybara? yes.  to Agile ventures? yes.

tansaku:
> still apply or do you want me to pull it out of the PR now? I'm confused by your last post about where you stand on it now.

Apologies for confusion.  Up to you.

freeranger:
Lets leave it in then.  Although the issue raised was about the visual appearance of long nicknames and so adding a tooltip is, strictly speaking, breaking
> 4) ensure the Pull Request corresponds to the letter of the ticket

I did ask in the discussion on the issue itself if I should add a tooltip and you said yes, so I'm going with your original answer ;)

diraulo:
hey @freeranger could you please have a look at the merge conflict and I'll pull this in.

tansaku:
@diraulo - the conflict is just with the changelog - and I think that's a good indicator we maybe should drop the changelog ... although I like the idea of having it there - it seems too much of a chore.

I resolved the conflict locally and pushed back up so you should be clear to merge the PR

diraulo:
Thanks, @tansaku. Will pull this in as soon as build's complete.

freeranger:
Thanks for fixing @tansaku yes it was a good idea at the time but with multiple PR's going on which end up on the same line of the changelog, it is a pain point.  I guess you could argue that if you want to see a change log then just look at the commit history in github....

PR #916, opened 2016-03-16
Allow event creator to specify related project

### LINKED ISSUE #796: Ability to connect scheduled events to a project
```
As an event planner,
In order to connect my event's hangout recording to my project,
I would like to set a project connection in the Event setup interface.
```
Right now, when an event is scheduled and started from the Event#show page, no `project_id` is being assigned to the `EventInstance` instance, thus it is not being shown on the `Videos` tab for the project. I would like that to happen. One idea is that we can allow who ever is planning an event to connect it to a project on the Event#create or #edit

tansaku:
fixes #796


diraulo:
conflicts here too @tansaku

tansaku:
fixed here too @diraulo

diraulo:
@tansaku this is just a partial fix to #796. Having the project on the event is not enough. I think this value should also be set to the `EventInstance` if we want to have recordings and others displayed on the project's page.

tansaku:
@diraulo yeah - we fix that in #921 - but it's like a trivial additional fix that might not even have been needed ...

PR #932, opened 2016-03-22
puffing billy config

### LINKED ISSUE #904: Network Sandboxing of our Cucumber Acceptance tests
Currently our Cucumber Acceptance tests are not completely sandboxed; at least not on the client side.

Ideally any tests that are running a headless browser with javascript enabled should be sandboxed from making any live network connections with something like [puffing billy](https://github.com/oesmith/puffing-billy) which will record and save any network activity for subsequent playback.

The server side version of puffing billy is [vcr](https://github.com/vcr/vcr) which records server side network activity for playback.  

The advantage of using vcr and puffing billy is that the recorded files (containing network actiivty that can be played back) are stored in a consistent location and can be easily updated by simply being deleted and having the tests be re-run, which avoids the need to manually edit custom webmock stubs.

The added advantages here should be that our tests run faster (don't have to wait for any network connections) and that they are more reliable (no failures due to failing network connections)

tansaku:
fixes #904


tansaku:
nice to see this is green - is it too early to hope this might fix our intermittent js issues ...?

PR #946, opened 2016-03-24
adds timeline import back into application.scss

### LINKED ISSUE #873: Styling issues on past scrums view
screenshot 2016-03-08 13 29 52

Also, investigate failure to load `scrums.js`

screenshot 2016-03-08 13 36 50

tansaku:
fixes #873

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -10,6 +10,8 @@
 @import "global/*";
 @import "layout/*";
 
+@import "timeline";
+
 @import "components/dropdown";
 @import "components/calendar-date";
 


tansaku:
@diraulo @tochman this should fix the scrum timeline view.  

I'm not completely happy with the fix.  It seems like we should do the timeline import in scrums.scss or similar, but I tried that and it didn't immediately.

So this is just a quick fix to see if pulling timeline back into application.scss has any knock on effects on the rest of the site - I'm not clear why @yggie pulled it out in the first place, but anyway, I'm not a scss/rails-asset expert, so perhaps one of you might have a better idea about what's the best way to fix this so the uneeded timeline styles are not being pulled in on every page ...

tansaku:
@diraulo did you have a chance to review this?

PR #948, opened 2016-04-04
makes slack service youtube notification include project slug

### LINKED ISSUE #947: Notification on Slack of Hangout creation
Since about 1 or 2 weeks ago no more notifications are shown in the Slack channels when we create a new PP session or a Standup.

Steps to reproduce the problem:
* Go to website one
* Create a PP session
* Press the notify to slack button

tansaku:
fixes #947


tansaku:
@diraulo hopefully this should be a pretty straightforward pull in

PR #969, opened 2016-04-07
trying to clean out our acceptance tests

tansaku:
fixes #806

PR #970, opened 2016-04-07
Update ruby and some gems

yggie:
- Updated Ruby to `2.3.0` and also the unlocked gems.
- Fixed failures as a result of the upgrade.
- Fixed a test order dependency bug in cucumber

View a cleaner diff [here](https://github.com/AgileVentures/WebsiteOne/commit/15a834263418cc53a185877dab008c061b73593f).


tansaku:
looks like progress @yggie - I guess we now need to run this many times on many machines to check how much of an impact this has on our intermittent failures ...

tansaku:
I'm getting a libv8 install failure on my dev machine trying to run bundle on your new branch ... but this has fixed it: http://stackoverflow.com/a/34851673/31672

also got this message:

```
Post-install message from acts-as-taggable-on:
When upgrading

Re-run the migrations generator

    rake acts_as_taggable_on_engine:install:migrations

This will create any new migrations and skip existing ones
```

tansaku:
cukes green for me locally, but then they generally are ...

yggie:
@tansaku I don’t think this will solve all of the intermittent failure issues, but this is a step forward…

tansaku:
oh definitely - your efforts very much appreciated here @yggie - at the very least we can see that the Ruby 2.3 upgrade isn't so painful.

do have a look at our cleanroom branch - maybe it's overkill, but our approach yesterday was to basically blow away all cuke config and start from scratch.  Ultimately I don't think we should be needing to switch js drivers on and off with special hooks - and we avoid that in the new config, so it should have the same advantage as your changes here.

The challenge with the cleanroom approach is that we now need to pull all the cukes back in one at a time to watch for overlaps, but it does have the advantage of a review of each and a complete sandboxing to avoid another source of randomness ...

diraulo:
1st, 2nd run --> Pass

3rd run
```shell
Failing Scenarios:
cucumber features/scrums.feature:19 # Scenario: Closing an existing video and opening a new one should update the player
``` 

4th run
```shell
Failing Scenarios:
cucumber features/hangout.feature:146 # Scenario: Infinite scroll on hangouts scroll down until no more hangouts
```

all next run (at least x4) --> Pass

yggie:
@diraulo to be honest, not sure if this is an improvement or not :sweat_smile:

diraulo:
@yggie, at least, this seems more stable than what I had after gems upgrades on my branch

tansaku:
@yggie I note that this runs quite clean for me locally, so those extra deprecation messages etc. and failures when I merged this into the cleanroom are likely artifacts of that merge ...  I did still get this warning, but that's a very simple fix I think:

```
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was #<ActiveRecord::RecordNotFound: User has not exposed his profile publicly>. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/tansaku/Documents/GitHub/AgileVentures/WebsiteOne/spec/controllers/users_controller_spec.rb:67:in `block (4 levels) in <top (requi
```

yggie:
@tansaku the warning has been resolved :smile:

tansaku:
thanks @yggie - @mtc2013 and I finished the cleanroom branch and I'm recommending to @diraulo to get that pulled in first since it was a marathon thing and I had trouble pulling the 2.3. into that ... sorry if that pushes more work into this PR that way, but I think strategically it's the right thing to do, and I'm very happy to help build the 2.3 upgrade on top of the cleanroom fix; as I know we still haven't conclusively fixed the intermittent js failures ...

yggie:
@tansaku no problem, this PR was never meant to address the intermittent failures, just simply an experiment to get it working with 2.3.0. How far have you guys gotten with the cleanroom fix?

tansaku:
@yggie all done https://github.com/AgileVentures/WebsiteOne/pull/969 - we've pulled everything back in and have things running with a much cleaner/simpler cucumber setup.  It hasn't completely fixed the issues, but things seem more stable.  I'd like to get it merged in and then pick up on the individual failures as they come up.  Now we've locked down all the network variability I'm keen to follow up on some of @freeranger's suggestions about adjusting the way that some of the HTML elements are selected ...

tansaku:
@yggie do you think it's worth trying to address the conflicts here or better start over?

yggie:
@tansaku I will try to `cherry-pick` the relevant changes across first, it might work out better that way

yggie:
ran into an unexpected issue… going to break this PR up into smaller pieces… https://github.com/AgileVentures/WebsiteOne/pull/1010 will be solely for upgrading the Ruby version, will keep this as a separate PR to update the gems…

PR #990, opened 2016-04-14
989 filter events by project ui

### LINKED ISSUE #989: UI for filtering by project
As a user
So that I can participate in events specific to the project I care about
I would like to filter the events list by project

mtc2013:
fixes #989

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,4 +1,3 @@
-
 class EventsController < ApplicationController
   before_action :authenticate_user!, except: [:index, :show]
   before_action :set_event, only: [:show, :edit, :update, :destroy, :update_only_url]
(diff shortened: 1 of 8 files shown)


diraulo:
thanks @mtc2013. A few comments

- Perhaps add an 'all' option in the select list to view all events again. Maybe this can be taken care of as part of a separate issue (if relevant at all)

- I think it would be better to make use of an inline form here:
screenshot 2016-04-18 11 32 24

see http://getbootstrap.com/css/#forms-inline

diraulo:
screenshot 2016-04-19 10 27 31

@mtc2013 @tansaku how about having the button on the right? feels more practical

tansaku:
maybe if we right align the text in the dropdown ... and thanks for those links to inline bootstrap forms - didn't see your comment until after we'd used the rows approach - the inline form look will probably be nicer ... I want to refactor some of the code as well.

tansaku:
hi @diraulo - we tried putting the button on the other side, but then we felt it needed a label, but then we couldn't get the label to vertical align nicely, so we ended up switching back ...

We did pull in the form-inline stuff and refactored the controller code ...

tansaku:
@diraulo updated based on input from kickoff meeting - hoping this is good to go now

PR #1027, opened 2016-05-05
1026 update airbrake config

### LINKED ISSUE #1026: Deployment to heroku breaks due to updated airbrake config
As of Airbrake 5, we now need to set a `PROJECT_ID` and `PROJECT_KEY` in the application configs.

diraulo:
Fixes #1026

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -19,7 +19,7 @@ WebsiteOne::Application.configure do
   # For large-scale production use, consider using a caching reverse proxy like nginx, varnish or squid.
   config.action_dispatch.rack_cache = true
 
-  config.serve_static_assets = true
+  config.serve_static_files = true
   config.static_cache_control = "public, max-age=31536000"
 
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -89,8 +89,7 @@
   config.log_formatter = ::Logger::Formatter.new
 
   # may be needed for integrating bootstrap with Heroku deployment
-  #config.cache_classes = true
-  #config.serve_static_assets = true
+  # config.cache_classes = true
```
tansaku: could these comments and commented code go off now?  bootstrap seems to work for us on Heroku ...?

tansaku:
looks good :shipit:

config/environments/production.rb
diraulo: Yeah I actually didn't pay close attention to the comment about bootstrap :)
I'll get rid of that 😉

PR #1036, opened 2016-05-18
event update redirection to show page

### LINKED ISSUE #935: saving an event after an edit should return you to the edit show page, not the list page
Returning to the list page can be very confusing, particularly if the edit you made removed the event from the upcoming list in view

touchps:
fix #935


touchps:
Acceptance test and controller specs both needed change.

PR #1044, opened 2016-05-19
adds support for premium plus

### LINKED ISSUE #1042: support premium plus subscription
support ability to pay for premium plus - to get pairing with AV mentor

tansaku:
fixes #1042

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -4,13 +4,12 @@ class ChargesController < ApplicationController
   end
 
   def create
-    # Amount in cents
-    @amount = 1000
+    @plan = params[:plan]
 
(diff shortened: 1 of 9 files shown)


tansaku:
This will fail because, I think, our current testing will not handle two stripe forms in one page.

There may be a fix for that in the test, or we may need to move them to separate pages, or have a single form with variable settings, or use a subscription upgrade option ...

tansaku:
there is a related issue about having the cached files stored in separate directories: https://github.com/oesmith/puffing-billy/issues/152

tansaku:
thinking about different url schemes, e.g.


# /subscriptions/new?plan=premium

# plan/premium/subscriptions/new

PR #1045, opened 2016-05-20
makes step do what was originally indented?  and updates feature to correspond to current functionality

### LINKED ISSUE #903: basic steps meaningless line
The following link in basic_steps `Then(/^I should( not be able to)? see a link to "([^"]*)" page for ([^"]*) "([^"]*)"$/)` is not actually an assertion and has no impact on test success failure

```
page.has_link?(action, href: url_for_title(action: action, controller: controller, title: title))
```

This should be turned into an expectation or removed, and sensibly we should review the role this step is playing in the related cucumber issues ...

tansaku:
fixes #903

### DIFF
--- a/features/sidebar.feature
+++ b/features/sidebar.feature
@@ -32,10 +32,10 @@ Feature: Sidebar navigation
 
   Scenario: Sidebar always shows the relevant information
     Given I am logged in
-    And I am on the "Show" page for document "Howto"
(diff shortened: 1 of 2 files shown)


diraulo:
@tansaku is this ready to be pulled in?

tansaku:
i think so

PR #1047, opened 2016-05-20
sponsor change from airpair to mentive

### LINKED ISSUE #1018: replace airpair with mentive on site
airpair descending while mentive is ascendant

touchps:
Fixes #1018 

I've left airpair.jpg in assets just in case. :)


tansaku:
great work @touchps - maybe ultimately we should replace the view spec with an acceptance test, but let's not worry about that in this pull request - we should get this deployed ASAP

PR #1062, opened 2016-05-25
c9 install docs

### LINKED ISSUE #1061: create c9 install documentation
We really need to help people have a smooth install process for C9

tansaku:
fixes #1061


tansaku:
@diraulo I think we have this working ...

PR #1065, opened 2016-05-25
Allow user to create event in their local timezone

### LINKED ISSUE #782: Allow users to create events in their local timezone
As a user 
So that I don’t need to convert my time into UTC
I’d like to enter the event time in my local-time.

tansaku:
fixes #782


tansaku:
As @mtc2013 was pointing out we'll need tests for editing the time after creation to ensure that time draw correct. We'll need JS to adjust the UTC times coming out of the DB to match the timeline of the user viewing the event ...

mtc2013:
@tansaku just pushed a change to load tzinfo labels into a select box

mtc2013:
@tansaku got it selecting the user's timezone to some degree based on IANA name with tzinfo and jstz libraries.

mtc2013:
Need to deal with bug where failed validation causes form to reload then the time drifts because the time is assumed to be in UTC whereas it is actually a user input in their local time.

mtc2013:
@diaraulo this is ready for code review

PR #1077, opened 2016-06-04
slightly refactored event model

SuperManEver:
moved code that retrieves event instances to EventInstance model. I think this is where it should resides


SuperManEver:
Not all cucumber tests are passing, but none of them is related to events subsystem. All rspec tests passing. I'm not 100% sure that failing cucumber tests are failed because of modifications I made.

.gitignore
```diff
@@ -71,3 +71,7 @@ Guardfile
 .vagrant/
 tags
 .byebug_history
+
+# testing cache
+/features/support/fixtures/req_cache
+/features/support/fixtures/cassettes
```
tansaku: these need to not be in .gitignore - otherwise our tests will spam 3rd party servers

tansaku:
thanks for this submission @SuperManEver - we have an intermittent failure affecting you here - I think ...  not sure needs investigation.

I'm sure that the event instances do need refactoring, but I haven't had time to review your suggestions as I've been working on fixes on event time zones and it's a crazy busy week and I'm not sure when I'll have time to review this.

Note that with AV premium you can get a guarantee of a full code review within 48 hours.

.gitignore
SuperManEver: are you suggesting to remove this lines?

tansaku: I am.  If you add those lines to gitignore then our acceptance tests will not be properly sandboxed by vcr and puffing billy:

* https://github.com/vcr/vcr
* https://github.com/oesmith/puffing-billy

We can't merge in with those lines present.

If you are getting stray files in those directories it means that you are experiencing network "leaks" when you are running the acceptance tests.

Ideally one would be investigating where those leaks are coming from, although that can be a complex topic.  It would be relatively easy for me to show you in a pairing session, or you can research the github links above.

diraulo:
@SuperManEver Here's the failing feature



This is most likely failing because you removed the file `Sign_up_for_premium_membership.yml`

tansaku:
@SuperManEver so we'll close this for the moment right? As we discussed in the retro I think we need to be cautious about refactorings away from those being driven by stakeholder needs

SuperManEver:
yes

PR #1087, opened 2016-06-08
Instructions to solve signup issues

nitishmadhukar:
Doc contains instructions to solve various signup issues
Currently has solutions for github public email


tansaku:
thanks @nitishmadhukar - good idea - we should also link this into docs homepage so that it can be discovered by navigation

nitishmadhukar:
@tansaku: My thought is once this doc is up in the master branch, we can link any user to this page when they are displayed with the Public Email error message after signup

tansaku:
@nitishmadhukar yes, that's a great idea - I was just thinking that we should also link to it from this page:

https://github.com/AgileVentures/WebsiteOne/blob/develop/docs/README.md

Which is an overview of what documentation is available, and you could include an update to that in this pull request ...

nitishmadhukar:
@tansaku updated README.md

PR #1099, opened 2016-06-13
Disable G+ signups while it doesn't work by hiding

### LINKED ISSUE #1095: Temporarily disable G+ signup
Having this on the site causes a lot of confusion when users try to sign up. We should remove this from site at least until #1063 is resolved

mtc2013:
@diraulo please review.

fixes #1095

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -4,8 +4,13 @@
   <% if resource.password_required? %>
       <div class="row">
         <% supported_third_parties.each do |provider| %>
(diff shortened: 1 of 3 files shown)


diraulo:
@tansaku @mtc2013 Perhaps out of the scope of this feature, but shouldn't we hide it on the sign_in page as well?

diraulo:
if not I'm happy to merge this 😉

mtc2013:
@diraulo @tansaku had said that we shouldn't prevent people from signing in who already have accounts.

PR #1100, opened 2016-06-13
Wraps more tests around the timezone detection during event creation and refactors out jstz in favour of moment js library

### LINKED ISSUE #1086: consolidate timezone javascript library usage
could replace jstv usage with moment usage.

Should probably increase test coverage in advance of refactoring ...

tansaku:
fixes #1086

More tests but could still use a workout manually to check we haven't missed anything.

Other refactorings that might follow this include:

* further clean up of select_time_zone....js file (breaking up methods, wrapping in object, doing in ember/react)
* using proper jasmine stubs 
* cleaning up helpers in event creation form
* loading js assets via npm
* further refactoring of event steps
* scenario outlines to replace replication over timezone checking for different event types


SuperManEver:
I think we need to encapsulate all the logic related to determining timezone into one object/module just to have a single source of truth about timezone and encapsulate all the stuff/logic related to this into this object. So, that we can safely switch to moment.js from jstz inside this object without affecting public interface

diraulo:
@tansaku update this branch with latest from `develop`

tansaku:
@diraulo done - btw - do you get permission to do that yourself through the github interface now? I just clicked a button ...

diraulo:
Yeah i noticed that this morning :) Will do that in future

PR #1102, opened 2016-06-13
Pushes changes from develop to staging

diraulo:


diraulo:
@tansaku any idea why we still get the following commit as part of this PR considering this code is already in production? 🤔

screenshot 2016-06-13 20 23 51

I think we should really review the flow we're currently using to push code from develop to the other branches

tansaku:
@diraulo I don't know - a good thing to add to the discussion for Friday ...

PR #1119, opened 2016-06-17
adds a hypothetical devops feature for slack

### LINKED ISSUE #931: Best approach for documenting devops?
@mtc2013 and I were testing on staging and we needed to set some env vars for managing the slack notifications including SLACK_NOTIFICATIONS and AGILE_BOT_URL and saying we thought it would be good to document this, but it was unclear where, other than to reflect they can be inferred from the code.

My big concern is about documentation going out of date, and I wondered if it made sense to have devops documentation in Cucumber like so:

``` gherkin
Feature: Slack Service
  As a site admin
  So that users can join in each others pairing sessions
  I would like them to be aware of each others activity

  Scenario: Slack Notifications
    Given that Slack Notifications are enabled  --> that the env var "SLACK_NOTIFICATIONS" is set to true
    And that the Agile Bot URL is set correctly --> that the env var "AGILE_BOT_URL" is set to "http://agilebot.herokuapp.com"
    When I start a hangout
    There should be an appropriate notification in slack
```

tansaku:
fixes #931

### DIFF
--- a/features/devops/slack_notifications.feature
+++ b/features/devops/slack_notifications.feature
@@ -1,0 +1,10 @@
+Feature: Slack Service
+  As a site admin
+  So that users can join in each others pairing sessions
+  I would like them to be aware of each others activity
+
+  Scenario: Slack Notifications
+    Given that Slack Notifications are enabled
+    And that the Agile Bot URL is set correctly
+    When I start a hangout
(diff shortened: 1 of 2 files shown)


tansaku:
I think this is a great overhaul approach but it's not practical under current resource constraints - will come back to it

PR #1120, opened 2016-06-17
1064 - instructions to signup with github public email

nitishmadhukar:
Provides link to GitHub page (instructions to sign up) in public email error message

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -151,7 +151,7 @@ class User < ActiveRecord::Base
     if email.blank? and not @omniauth_provider.nil?
       errors.delete(:password)
       errors.delete(:email)
-      errors.add(:base, "Your #{@omniauth_provider.capitalize} account needs to have a public email address for sign up")
+      errors.add(:base,  I18n.t('error_messages.public_email', provider: @omniauth_provider.capitalize))
     end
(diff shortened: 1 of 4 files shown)


tansaku:
great work @nitishmadhukar - well done incorporating all the feedback on the last pull request - this is great stuff and ready to be pulled in.

There is something else to look at which is the readability of 

```gherkin
  @omniauth-without-email
  Scenario: User signs up with a GitHub account having no public email (sad path)
    Given I am on the "registration" page
    When I sign up with GitHub
    Then I should see link for instructions to sign up
```
I think it's  a little odd that we have the "without email" aspect in a tag.  Now that's not something you introduced - so doesn't need changed before we pull this in and I think we should pull this in ASAP @diraulo 

but we might open a refactoring ticket to adjust this scenario to something like:


```gherkin
  Scenario: User signs up with a GitHub account having no public email (sad path)
    Given I am a Github user with a private email 
    When I sign up with GitHub
    Then I should see a link to detailed instructions to make my GitHub email public
```

diraulo:
Hi @nitishmadhukar please rebase your branch with `develop` and update this PR so I can merge this. That should fix the merge conflicts

PR #1143, opened 2016-06-23
Fixes problem with incoming API for hangoutconnection where similar submissions would not update updated_at

### LINKED ISSUE #1073: Indicate if event is live when the event is actually still in progress
It appears the indication of events being live only appears for a moment and then goes away... We should in principle keep that until the event ends.

tansaku:
fixes #1073 

and lots of great exciting test coverage for this issue


diraulo:
@tansaku please have a look at this merge conflict

PR #1181, opened 2016-06-30
934 event edit state

### LINKED ISSUE #934: when event edit (on existing event) fails due to repeat end being blank, state about days is lost
Need to review maintaining appropriate state across validation failures.

Arepo:
We weren't too happy about this fix, but couldn't think of any approach that seemed obviously better.

fixes #934


tansaku:
great work @Arepo - I love a quick basic fix!

@mtc2013 were you going to open a chore for the more detailed refactoring you suggested?

@diraulo if you're happy let's merge this in.

PR #1202, opened 2016-07-06
Adds mentor page

### LINKED ISSUE #1123: mentors page
Would be good to identify the AV mentors and have a place where it was easy to browse their profiles:

http://agileventures.org/mentors perhaps?

```gherkin
As a prospective premium(plus) member
So I can see who I would be getting support from
I'd like to know more about AV mentors background
```

mtc2013:
Fixes #1123


tansaku:
@diraulo looking at the codeclimate issue, I would say it is acceptable for now - this should be released urgently as part of our business model.

I'll open a refactoring chore for the issue - there's some drying our we could do between types of filtering

PR #1230, opened 2016-07-13
1130_feature for unlinking github account

### LINKED ISSUE #1130: user unable to change github user name
User reported:

> I changed my Github username after I created my AgileVentures ID, and I don't see a way to update my new Github username on my AgileVentures account page. When I click on "Remove Github" (so that I connect it again to the new one), it gives a red flag that says, "Bad Idea."

touchps:
Fixes #1130

I've created a feature which looks fine but failing at step `When i click "Remove Github"`. Checking contents of the page in capybara also reveals that it contains like "Remove Github". 
Any pointers appreciated.


features/users/user_management.feature
```diff
@@ -58,4 +58,14 @@ Feature: Create and maintain projects
...
-    Then I should see "by Anonymous"
\ No newline at end of file
+    Then I should see "by Anonymous"
+
+  @omniauth
+  Scenario: Unlink my github link from my profile
+    Given I have a GitHub profile with username "tochman"
+    And I have authentication enabled with my github username
+    And I am on my "Edit Profile" page
+    When I click "Remove Github"
```
tansaku: as we saw in hangout - looks like this should be "GitHub"

tansaku:
thanks @touchps - want to update the description of the PR now that we have fixed that error?

app/controllers/authentications_controller.rb
```diff
@@ -35,7 +35,12 @@ def destroy
         # Bryan: TESTED
         flash[:alert] = 'Bad idea!'
       elsif @authentication.destroy
-        flash[:notice] = 'Successfully removed profile.'
+	user = User.find(current_user.id)
+	if user.update_attributes(github_profile_url: nil)
+          flash[:notice] = 'Successfully removed profile.'
+	else
+          flash[:notice] = 'Github profile url could not be removed.'
```
tansaku: should we say why? e.g. "could not be removed because no other authentication mechanism is in place - please add G+ authentication in order to remove your GitHub authentication"

features/users/user_management.feature
```diff
@@ -58,4 +58,15 @@ Feature: Create and maintain projects
       | Documentation | My documentation | hello world |
     When I delete my profile
     And I am on the "Show" page for project "hello world"
-    Then I should see "by Anonymous"
\ No newline at end of file
+    Then I should see "by Anonymous"
+
+  @omniauth
+  Scenario: Unlink my github link from my profile
```
tansaku: should we also have a scenario for the sad path?

touchps:
This change add extra code for clearing up the github_profile_url from user attributes in case user has already authenticated using github and trying to unlink github account. This will only work if user has a backup authentication strategy in place (password or gplus).
Summary of changes;-

- added cucumber feature and supported step definition for unlinking github account
- added rspec test for authentications controller to test if guthub_profile_url gets cleared
- actual code change in the authentications controller destroy action.
- change in other authentication controller spec for other tests on destroy action. These changes include stubbing due to additional method call in destroy (detected via running all rspec tests at end)

features/users/user_management.feature
touchps: make sense. Although there is controller level spec to test that. But its good to have a acceptance level test in place.

touchps:
Sad case added for unlinking github account. Error message is changed from `Bad Idea` to `Failed to unlink GitHub. Please use another provider for login or reset password.`

The changes look ok for now, but i'm thinking of some more refactoring. There are some paths that doesn't seem to be reachable using acceptance test, probably due to change in logic. One way to cover those paths is to add deliberate stub failure in unit tests. Will discuss this in next scrum.

Changes as per last review are complete.

app/controllers/authentications_controller.rb
```diff
@@ -33,9 +33,14 @@ def destroy
     if @authentication
       if current_user.authentications.count == 1 and current_user.encrypted_password.blank?
         # Bryan: TESTED
-        flash[:alert] = 'Bad idea!'
+        flash[:alert] = 'Failed to unlink GitHub. Please use another provider for login or reset password.'
```
tansaku: this is great - I might chance to "Failed to unlink GitHub.  To unlink GitHub please first connect your account to another provider (e.g. Google+) or reset your password to create email authentication.", but even with your text above this is a great improvement ... I mean we should double check that resetting password is an option for the user if they only ever signed up via a 3rd party provider.

The key thing to thing about here @touchps is that while we internally might understand what a "provider" is, there's no reason that our users will be thinking in terms of that word.   That's the problem with so many error messages - they are constructed from the developers point of view and using the developers vocabulary; can be very confusing for user ...

app/controllers/authentications_controller.rb
tansaku: @touchps following up on this - was this the flash part that wasn't reached by the tests?  Do we even know what might cause a failure here?

touchps: @tansaku 
line 45: wasn't reached. It covers the case when update_attributes fails to set github_profile_url as nil. This scenario seems rather not quite predictable in acceptance tests

line 48: wasn't reached. It covers sad case when github_profile_url is present, but no github authentication. I'm not sure how we can have the scenario can happen. 

Being unsure on both of these, i thought removing them might be a risk.

tansaku: removing the code as a risk? yes indeed - but can we create new tests that reach these locations?  You think that's tricky?

tansaku:
@touchps and I discussed this in the "Kent Beck" scrum yesterday - although there could be some more tests to wrap some of the existing functionality, the functionality that @touchps is adding is fully tested, so this could be merged to finally fix the issue originally reported by the user.

We just noted some quick refactoring that could be done here - which I pushed up - that just needs merged in this pull request and I think we'll be good to go ...

tansaku:
as well as fixing merge conflicts

touchps:
@tansaku 
Did some more analysis. The case for `authentication not found` is not reachable (not even in terms of prev legacy code). The unlinking option only appears if a authentication is in place. This means you won't be able to login via github if github_profile_url is present and authentication not in place(unless you sign in using github provider again).

Considering this, I've removed unnecessary logic to make it simple. No new test cases added, all tests passing.

tansaku:
@touchps good show - probably time to get this deployed and see what happens - @diraulo want to have a once over?

diraulo:
@tansaku no comments on this. I think it's good to go

PR #1236, opened 2016-07-24
error message for failed github unlinking

touchps:
Adds more feedback in error message. 

This separates out the message change for issue #1230 and should be merged first before #1130 which focuses on resolving github unlinking.

### DIFF
--- a/app/controllers/authentications_controller.rb
+++ b/app/controllers/authentications_controller.rb
@@ -33,7 +33,7 @@ class AuthenticationsController < ApplicationController
     if @authentication
       if current_user.authentications.count == 1 and current_user.encrypted_password.blank?
         # Bryan: TESTED
-        flash[:alert] = 'Bad idea!'
+        flash[:alert] = 'Failed to unlink GitHub. Please use another provider for login or reset password.'
       elsif @authentication.destroy
         flash[:notice] = 'Successfully removed profile.'
       else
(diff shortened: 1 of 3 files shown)


tansaku:
@touchps ah fantastic - yes, the minimal PR - love 'em!

app/controllers/authentications_controller.rb
```diff
@@ -33,7 +33,7 @@ def destroy
     if @authentication
       if current_user.authentications.count == 1 and current_user.encrypted_password.blank?
         # Bryan: TESTED
-        flash[:alert] = 'Bad idea!'
+        flash[:alert] = 'Failed to unlink GitHub. Please use another provider for login or reset password.'
```
tansaku: this is great - I might chance to "Failed to unlink GitHub. To unlink GitHub please first connect your account to another provider (e.g. Google+) or reset your password to create email authentication.", but even with your text above this is a great improvement ... I mean we should double check that resetting password is an option for the user if they only ever signed up via a 3rd party provider.

The key thing to thing about here @touchps is that while we internally might understand what a "provider" is, there's no reason that our users will be thinking in terms of that word. That's the problem with so many error messages - they are constructed from the developers point of view and using the developers vocabulary; can be very confusing for user ...

tansaku:
@diraulo this is ready for your review I think - have a look at my comment.  Even if we push out as is, this a big improvement on the "bad idea" message we had originally ...

app/controllers/authentications_controller.rb
touchps: The detailed message did came in my mind, however i wanted to keep message size as minimum as possible with still delivering the intended message. Keyword `provider` was used to keep it generic, just in case if we decide to add one more provider.

diraulo:
Hi @touchps, in what order should these PRs be merged? This one first then #1230 or the other way around?

touchps:
This one should me merged first. Minor analysis pending for #1230. Will update 1230 once ready for merge.

diraulo:
great thanks @touchps

PR #1246, opened 2016-08-27
Proposal for issues and pull requests templates

diraulo:
Github now has supports for issues and pull request templates. I thought it'd be nice to have the on this project to allow contributors to properly document their PRs or issues. 

These templates were generated with the help of https://www.talater.com/open-source-templates. Some items might not necessary apply to our current project. @tansaku @mtc2013 Let me know what you think of them. :)


tansaku:
Great idea @diraulo - I think templates are awesome - will take a detailed look through these specific ones ASAP

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+4.
+
+## Context
+<!--- How has this issue affected you? What are you trying to accomplish? -->
+<!--- Providing context helps us come up with a solution that is most useful in the real world -->
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+* Version used:
+* Environment name and version (e.g. Chrome 39, node.js 5.4):
```
tansaku: could change this from node.js to ruby/rails version?

tansaku:
@diraulo looked at these in more detail and they look good!

.github/ISSUE_TEMPLATE.md
diraulo: I think in this context we could make this about the environment in which they're running the application. We already have a fixed ruby/rails version here.

tansaku: right - so we should just delete that mention of node.js then, since we don't use it and having it there might be a little confusing and then I think we can merge these ...

tansaku: do I have push access to your repo?  if so I could do that for you ... or feel free to make these branches on the main repo (that's what @mtc2013 and I do) and then we don't have to worry about access issues.  Of course if you're keen to keep a repo we can't push to then we totally respect that too :-)

diraulo: I'm still used to my old workflow of creating branches on my own fork. But I think I'll just create new branches from the main repo in future ;)

diraulo:
@tansaku I think this is ready to go. I removed the mention of nodejs in the issue template

PR #1251, opened 2016-09-01
Allows user to update card details

### LINKED ISSUE #1241: Enable premium users to update card details
As the number of premium users increases this is going to become more and more important.  Some related links:

* https://stripe.com/docs/recipes/updating-customer-cards
* http://stackoverflow.com/questions/20065939/change-credit-card-information-stripe

tansaku:
fixes #1241


app/views/charges/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_tag charge_path('tansaku'), method: :put do %>
```
tansaku: @mtc2013 we should probably remove this hard coding

app/views/charges/edit.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  data-key="<%= Rails.configuration.stripe[:publishable_key] %>"
+  data-name="Agile Ventures"
+  data-panel-label="Update Card Details"
+  data-label="Update Card Details"
+  data-allow-remember-me=false
+  data-locale="auto">
+  </script>
+<% end %>
+
+<!--data-image="/path/to/your/logo.png"-->
```
tansaku: should remove this or get images sorted

tansaku:
hmm, and I was wondering if we needed to make the cukes more imperative ...

diraulo:
@tansaku is this ready to go? I need to update my card number 😉

tansaku:
@diraulo yes, this is good to deploy - we will have to do a little data migration before you can use it though

PR #1254, opened 2016-09-02
utc date removed from show page

touchps:
submitting the quick, ruthless and minimal change.


tansaku:
@touchps super awesome - so out of interest, what are the other options under consideration?

1) add a UTC date?
2) add an x hours till event?
3) ...?

tansaku:
> > 1) add a UTC date? 

> Not sure it would be of much help.

This would be a way of keeping the UTC time in, but avoiding any inconsistency.

> > 2) add an x hours till event? 

> Good only for meetings within 24-48 hrs.

Yes if only hours specified

> 3) relative time like 3 hrs/3 days/1 week etc.

> Last one feels most friendly and suitable too.

Yes, so having things like 3 days and 1 hour until meeting seems useful ...

tansaku:
@touchps @sampritipanda and I talked about this in the Kent Beck scrum and agreed that adding a counter could be addressed in another ticket

tansaku:
@diraulo I guess this is good to go - we didn't have any user complain about this specifically I think.  I think it was just @mtc2013 noticing a potential inconsistency, but I guess good to remove it and then we can start work on a separate tickets to continue improving the view here ...

PR #1263, opened 2016-09-14
improves layout landing page mobile

### LINKED ISSUE #1249: main page on iOS alignment looks pretty bad
Is it just me of should that let's code rounded box be centered against the background?

tansaku:
fixes #1249


diraulo:
@tansaku made some small changes to your branch to fix the alignment of those elements on mobile view.

screenshot 2016-09-16 14 03 03

Issue #1267, opened 2016-09-15
Failing Charge Acceptance test


Maroo-b:
Error message in cucumber test: features/charge_activity.feature
## Current Behavior

Error message **Unable to find field "Email" (Capybara::ElementNotFound)** in step **When I fill in appropriate card details for premium** (for all steps related to filling stripe popup)

For debugging I used save_and_open_page step and I find the following error in the page:

```
{
  "error": {
    "type": "invalid_request_error",
    "message": "Requested URL not found"
  }
}
```

tansaku:
thanks @Maroo-b feel free to throw in the full stack trace next time :-)  great to see the specific line numbers ... identify the particular scenario that fails - although perhaps all are failing.

Also can you do `git status` and show us what files are being generated?

Maroo-b:
@tansaku here is the full stack

**Scenario: Sign up for premium membership**
`When I fill in appropriate card details for premium     # features/step_definitions/charge_steps.rb:1
      Unable to find field "Email" (Capybara::ElementNotFound)
      ./features/step_definitions/charge_steps.rb:6:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/charge_steps.rb:5:in `/^I fill in appropriate card details for premium(?: for user with email "([^"]*)")?$/'
`

**Scenario: Sign up for premium plus membership**

`When I fill in appropriate card details for premium plus    # features/step_definitions/charge_steps.rb:29
      Unable to find field "Email" (Capybara::ElementNotFound)
      ./features/step_definitions/charge_steps.rb:32:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/charge_steps.rb:31:in `/^I fill in appropriate card details for premium plus$/'
      features/charge_activity.feature:32:in `When I fill in appropriate card details for premium plus'
`

**Scenario: User decides to change membership**

`When I fill in updated card details for premium for user with email "tansaku+stripe@gmail.com"                  # features/step_definitions/charge_steps.rb:15
      Unable to find field "Email" (Capybara::ElementNotFound)
      ./features/step_definitions/charge_steps.rb:20:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/charge_steps.rb:19:in `/^I fill in updated card details for premium(?: for user with email "([^"]*)")?$/'
      features/charge_activity.feature:40:in `When I fill in updated card details for premium for user with email "tansaku+stripe@gmail.com"'
`

--------------- **Git Status results** -------------------

[...]

PR #1273, opened 2016-09-16
1267 Edit stripe locale config

Maroo-b:
Addresses: https://github.com/AgileVentures/WebsiteOne/issues/1267

I changed locale setting of stripe from **auto** to **en-US**. So developers can work on project without need of tweaking cassette files and also to prevent displaying the popup in different languages for users.

### DIFF
--- a/app/views/charges/edit.html.erb
+++ b/app/views/charges/edit.html.erb
@@ -6,6 +6,6 @@
   data-panel-label="Update Card Details"
   data-label="Update Card Details"
   data-allow-remember-me=false
-  data-locale="auto">
+  data-locale="en-US">
   </script>
 <% end %>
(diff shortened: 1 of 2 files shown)


Maroo-b:
@tansaku the build fail is due to JS issue on features/projects/edit_project.feature 
is it frequent that you face this issue ?

tansaku:
@Maroo-b yes - pretty common - I have these cucumber tags to start removing them - although maybe we should try that repeating thing that @johnnymo87 suggested for LS - in the meantime I will re-start the build ...

tansaku:
hmm, that addresses when we merged didn't seem to close the related ticket ...

also, looking at this now I see that the data-locale is now hard coded to "en_US" which doesn't seem ideal.  @Maroo-b - wouldn't it be better to have some call to an env_var or something that we can stub?

Don't we now risk that customers in different locales will see en_US when they might prefer/expect their own locale?

PR #1275, opened 2016-09-19
experimental

tansaku:
this change would make our develop server mirror the settings of our production server better ...


tansaku:
@diraulo I think this should probably be merged if you're happy with it

tansaku:
@diraulo I need to do some testing with this - I'm going to merge - we can revert if a problem

PR #1276, opened 2016-09-19
1172 remove nearest timezone

mtc2013:
wip


tansaku:
@mtc2013 since we didn't see any improvement here and the site still seems to run, I guess we should close this for now?

PR #1279, opened 2016-09-20
Fix heroku deploy by adding bower to deps

mtc2013:
Pulling this in to clear blocking issue.

### DIFF
--- a/app.json
+++ b/app.json
@@ -52,6 +52,9 @@
   },
   "buildpacks": [
     {
+      "url": "https://github.com/heroku/heroku-buildpack-nodejs"
+    },
+    {
       "url": "https://github.com/heroku/heroku-buildpack-ruby"
     }
   ]
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "doc": "docs"
   },
   "scripts": {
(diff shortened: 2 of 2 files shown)


diraulo:
@mtc2013 deploy of review app failed with this:

```shell
Sprockets::FileNotFound: couldn't find file 'nprogress' with type 'application/javascript'
```

mtc2013:
@diraulo thanks.  i think it has to do with buildpacks not being set right and in the right order on the opened pr instances, ie develop-pr-1279.  I did this to the main develop server:

heroku buildpacks:clear
heroku buildpacks:set heroku/nodejs
heroku buildpacks:add heroku/ruby --index 2

I think this will get it past that error because it will run nodejs buildpack first.

mtc2013:
Perhaps this can best be integrated into the app.json somehow.

mtc2013:
@diraulo just pushed a change to app.json.  Hope that will clear things up.

mtc2013:
pushed another change that is probably more sensible.  preinstall hook to put in bower.

mtc2013:
@diraulo @tansaku successfully deployed. yah!  had to ultimately add bower to our own dependencies.

PR #1311, opened 2016-09-28
upgrades to ruby 2.3.1

### LINKED ISSUE #1310: upgrade to ruby 2.3.1
would be good

tansaku:
fixes #1310


tansaku:
@diraulo I guess this won't deploy until we specify ruby 2.3.1 on the CI?

diraulo:
@tansaku I think you need to run `bundle` to update the `Gemfile.lock` it still has the following in it. 
```ruby
RUBY VERSION
   ruby 2.3.0p0
```
That's probably the reason why Heroku is still trying to deploy using `2.3.0`

tansaku:
@diraulo I did do that already - when I do it I'm not getting any changes to Gemfile.lock for some reason ...

tansaku:
https://github.com/bundler/bundler/issues/4595

PR #1312, opened 2016-09-29
1291 improve telemetry from hangout

### LINKED ISSUE #1291: improve hangout telemetry
currently we only know who appeared in a hangout, but not when they arrived and left

mtc2013:
Fixes #1291


tansaku:
@diraulo this is ready for review - the codeclimate issue is on a method we've renamed - GPA is not changing - we think that can be ignored as we'll do a more thorough refactoring on that later

tansaku:
@diraulo I think this one could be merged - what do you think?

diraulo:
@tansaku was going to merge this weekend after I updated the PR, but got distracted by something else and forgot 😉

PR #1323, opened 2016-10-06
Allows premium users to upgrade to premium plus plans

### LINKED ISSUE #1303: allow premium users to upgrade to premium plus
this would require a new data model

tansaku:
It involves schema migration and has some Demeter violations that we should refactor out.
It will also involves data migration - I guess we should make a rake task for that.  I guess the critical learning from the last outing is don't delete old parts of the schema until we are sure everything works ...

fixes #1303


tansaku:
@mtc2013 so I got the tests passing here - but I think we need to review the process of data migration to the new structure before things can be deployed - does that require a rake task perhaps?  or better to work by hand for the 30 or so premium members we have currently ...

tansaku:
failure on deploying from PR is this:

Your role collab on the team neurogrid-ltd is not allowed to perform that action.

PR #1325, opened 2016-10-09
1324 hire me button removed for the logged in user 

### LINKED ISSUE #1324: "Hire me" link is visible on a users profile page when that user is logged in
## Current Behavior
"Hire me" link is visible on a users profile page when that user is logged in.  The link pops up a form that upon filling out, sends an email to the user.  It appears to me that this feature is meant to allow for people to contact users with job opportunities.  
image
## Impact
The presence of the link for a logged in user makes the "Hire me" feature confusing at first glance.  
## Expected Behavior
When the user is logged in, they should not see the "hire me" link on their own profile page.

c1505:
Fixes #1324 
This change removes the "Hire me" button from the logged in user's profile page.  The button is still visible to anyone except for the current user.

Removed ```@javascript``` from the cucumber tests related to this change.  With it in there, the ```sign out``` step fails with the error:  
```
undefined method `submit' for #<Capybara::Poltergeist::Driver:0x007f8f12ec4110> (NoMethodError)
      ./features/step_definitions/user_steps.rb:49:in `/^I sign out$/'
      features/users/profile_privacy.feature:102:in `And I sign out'
```

This is my first pull request on the project.  Any feedback or changes needed are welcome :)

### DIFF
--- a/app/views/users/profile/_summary.html.erb
+++ b/app/views/users/profile/_summary.html.erb
@@ -39,7 +39,7 @@
         </li>
     <% end %>
 
-    <% if presenter.display_hire_me %>
+    <% if presenter.display_hire_me && current_user != presenter.user %>
         <li>
           <%= link_to 'Hire me', user_path(presenter.user),
                       {remote: true, class: 'user-profile-btn',
--- a/features/users/profile_privacy.feature
+++ b/features/users/profile_privacy.feature
@@ -59,6 +59,7 @@ Feature: As a site user
     And "Display email" should not be checked
     When I set my email to be public
     And I click "Update"
+    And I am on my "Profile" page
     Then I should see my email
 
   @javascript
@@ -74,19 +75,20 @@ Feature: As a site user
 
   Scenario: Hire Me button should be private by default
     Given I am logged in
+    And I sign out
     And I am on my "Profile" page
     Then I should not see button "Hire me"
 
-  @javascript
   Scenario: Should be able to make my Hire Me button public
     Given I am logged in
     And I am on my "Edit Profile" page
     And "Display Hire Me" should not be checked
     When I set my Hire Me to be public
     And I click "Update"
+    And I sign out
+    And I am on my "Profile" page
     Then I should see button "Hire me"
 
-  @javascript
   Scenario: Should be able to make my Hire Me button private again
     Given I am logged in
     And My hire me was set to public
@@ -94,7 +96,16 @@ Feature: As a site user
     Then "Display Hire Me" should be checked
     When I set my Hire Me to be private
     And I click "Update"
+    And I sign out
     And I am on my "Profile" page
+    Then I should not see button "Hire me"
+
+  Scenario: Should not be able to see Hire Me button when logged in
+    Given I am logged in
+    And I am on my "Edit Profile" page
+    And "Display Hire Me" should not be checked
+    When I set my Hire Me to be public
+    And I click "Update"
     Then I should not see button "Hire me"
 
     # Bryan: To be added back later in another story


tansaku:
this looks great @c1505 !  I just wonder if we have an existing helper method to check if the user is on their own profile page, e.g. on_own_profile_page?  or similar.  If not maybe we should have - either way great work!

diraulo:
hmmm.. wonder why we don't have a review app on this PR 🤔

tansaku:
@diraulo I think it might be because we moved website-develop to a personal account to stop incurring heroku charges - I think the same thing happened on localsupport, where I'm getting deploy fails on develop - I think @joaopapereira encountered the same on metplus - there may be some tokens we need to regenerate or something ...

c1505:
@tansaku a on_own_profile? method seems like a good idea.  I could make that change if you guys prefer or at least use ```presenter.user``` rather than ```@user``` for consistency with the rest of the partial.

tansaku:
@c1505 sounds good.   I'm not 100% comfortable (yet) with the presenter pattern, but we should probably go with the grain here.

Perhaps it makes sense to add `on_own_profile?` to the presenter class ... hmm ... I think if we are using the presenter then where possible we should avoid exposing the user object directly via `presenter.user`.  Just thinking out loud here.     So maybe `presenter.on_own_profile?`

I'm also thinking about line 70 where @mtc2013 and I recently added ` <% if current_user && current_user == presenter.user %>`  ...

Maybe we should be able to ask:

```
presenter.matches_the_current_user?
```

because the presenter is wrapping a user object, so the simplest thing to ask is, is your user the logged in user ...

There are lots of alternatives, and we probably shouldn't spend too much time on this - hah - I'm generating several alts in my brain:

```
presenter.is_the_current_user?
presenter.wraps_the_current_user?
presenter.matches_the_current_user?
presenter.holds_the_current_user?

presenter.your_user_currently_logged_in?
presenter.user_is_viewing_this_page?
```

I think what we want is something that reads smoothly in the partial and also actually exposes what's happening under the hood at this point, for reference I imagine the method being:

```
def user_is_viewing_this_page?
  return false unless current_user
  user == current_user
end
```

ooh, and error checking there for confidence ... let me know what you think

c1505:
I don't think I have access to current_user in the presenter so I might have to pass it in or do something like this:
```
class ApplicationController < ActionController::Base
  prepend_before_filter { @@current_controller = self }
end

class YourPresenter
  def current_user
    ApplicationController.current_controller.current_user
  end
end
```
http://stackoverflow.com/questions/6998830/accessing-rails-helpers-from-a-presenter-module

I would lean towards just passing it in or keeping the logic in the view rather than making those changes.

Is the ``'return false unless current_user''' necessary?  If there is no current_user, then won't it always be false?  

I might be discussing too much for a fairly basic change.  Just trying to learn.  Let me know your thoughts.

tansaku:
oh yes, you're right - doh! no need to check for nil there - I still had my head in some code where we keep calling methods on current_user - forget that.

to get a cleaner solution to current_user in the presenter it seems like there's a new gem we could use: https://github.com/drapergem/draper

That might be worth investigating, but perhaps that should go in another PR/ticket.

I think we want to move relatively quickly to close this one out now, you could use that hack from SO if you like, or not preferring something

```
presenter.user_same_as? current_user

def user_same_as? another_user
  user == another_user
end
```

to close out fast get this merged in and then create some investigation tickets to explore options - then you might work on those, or not.  I think it's good to get PRs merged relatively fast, and give people a sense of completion.  And also the opportunity to switch task if they want to :-)

c1505:
I just decided to follow along with what is there and go with ```presenter.user```.  Potentially will investigate the gem later.

tansaku:
sounds good @c1505 we'll open some refactoring tickets

Issue #1245, opened 2016-08-26
link karma number to breakdown of karma on user profile page


tansaku:
Would be good to be able to go from individual karma numbers on users index page to breakdown of karma on profile ... might be a little tricky due to currently having hyperlink on whole thing

Maroo-b:
@tansaku I suggest to edit the hyperlink to user profile by setting it on user name and on his avatar only. The reason behind this even the whole block is clickable a user will try to point to the user name because it looks like a link.

tansaku:
good idea @Maroo-b !

Maroo-b:
@tansaku can you please provide me a link to breakdown of karma on profile is it the activity tab on profile ?

tansaku:
@Maroo-b it is the activity tab on the profile

Maroo-b:
@tansaku I was experimenting with development environment, and I noticed that it's possible to have users without activity tab in their profiles. So I think about two alternatives here:
1- To add link to karma number and in case the user doesn't have an Activity tab the system display the profile w a flash message "The user does not have Activity record yet" for example.

2- To add link only to users who have activity tab only. 

I prefer the first alternative because it keeps the consistency of the system and avoid making users guess why the link on Karma number isn't always present.

What do you think ?

PR #1328, opened 2016-10-11
1245 link karma number to breakdown of karma on user profile page

Maroo-b:
Addresses: https://github.com/AgileVentures/WebsiteOne/issues/1245

Changes description:

* link to user profile from list of user is accessible via clicking on user full name or avatar
* add link to user activity profile tab via karma number
* add verification method that displays a flash notice if user doesn't contributed yet
* add helpers method to handle active class changes for user tabs


tansaku:
Thanks for this @Maroo-b - have got it working locally - nice work.

One thing is that we lose the effect of having the user profile go orange when we mouse over ... that's not the end of the world, but I thought it was a nice feature that highlighted to the user the clickability of the profile.  What do you think?

Maroo-b:
@tansaku the mouse over effect was related to  wrapping the user profile with a link tag, it's possible to keep the effect but it will lead to user confusion because he will think that the item is clickable.

I tried to add underline style but I opt for keeping the default style used for links, (UI consistency stuff and the user already learnt that the orange text is a link)

Maroo-b:
@tansaku I was thinking about removing link coloring from Karma number (keep only the pointer cursor), any feedback ?

tansaku:
@Maroo-b we definitely want consistency - but the whole area in your pull still seems clickable - so not so misleading to change color.

As regards removing colour from karma link I would keep that colour for consistency no?

tansaku:
@Maroo-b needs sync to develop too

Maroo-b:
@tansaku I updated the PR to sync it with develop and remove additional files added by mistake.

I added the hover effect on user info block. After testing it, I agree that keeping this effect will be better.

tansaku:
@Maroo-b that looks great - just what the doctor ordered!

app/controllers/users_controller.rb
```diff
@@ -108,4 +109,14 @@ def set_timezone_offset_range(filter_params)
...
+
+  def set_activity_tab(param)
+    if param.present?
+      @param_tab = param
+      unless UserPresenter.new(@user).contributed?
+        @param_tab = nil
+        flash.now[:notice] = 'User does not have activity log'
+      end
+    end
+  end
```
tansaku: Just thinking you could shorten this up with 

```rb
def set_activity_tab(param)
  return unless param.present?
  @param_tab = param
   unless UserPresenter.new(@user).contributed?
     @param_tab = nil
     flash.now[:notice] = 'User does not have activity log'
  end
end
```

tansaku:
Added some comments for a refactoring
also I wonder why we have helper specs and controllers specs if they are not testing additional code paths that the acceptance tests don't catch?  What do you think?

Maroo-b:
@tansaku the controller spec is check the different cases:
* user with contributions
* user without contributions

The acceptance test is verifying the happy path only.
Do you think that I should remove the helper specs ?

Maroo-b:
@tansaku I added the refactor. Thanks.

tansaku:
@Maroo-b thanks for the explanation - I think that even with the two code paths all the use cases are covered by the cucumber tests - I might be wrong.  What I worry is that the helper and and controller tests bind too tightly to a particular implementation of the desired behaviour, making it more difficult to change in future.

Also I think the cucumber generally does a better job of exposing the high level intention.  I love good unit tests but I like them on services, models and things that might get moved into separate gems where they can really benefit from re-use.

That said I don't mean to stop this from being pulled in - it's great functionality that we really need.  I can see a couple of spelling issues - I'll try and push in a few changes soon.

Maroo-b:
@tansaku I deleted the helper test and reset the changes for users_controller_spec.rb. Also I added the an acceptance of visiting a user profile via karma link when the user didn't contribute.

It makes sense to follow the project style.

tansaku:
@Maroo-b thanks so much for this - note I corrected the syntax to link to the ticket like so `fixes #1245`

PR #1351, opened 2016-10-20
Refactor same-user checking logic into presenter

Arepo:
After discussion with Sam I've made the minimal change here, but would like to refactor the Presenter class to expect a View object and call methods on that, rather than either including or passing in bits of view logic.

#### Checklist:


tansaku:
looks great @Arepo - awesome smallness!

app/views/users/profile/_summary.html.erb
```diff
@@ -67,7 +67,7 @@
                          type: 'button' %></p>
         </li>
     <% end %>
-    <% if current_user && current_user == presenter.user %>
+    <% if current_user && presenter.user_same_as?(current_user) %>
```
tansaku: could we even pull the nil check in there into the user_same_as?

app/views/users/profile/_summary.html.erb
```diff
@@ -39,7 +39,7 @@
         </li>
     <% end %>
 
-    <% if presenter.display_hire_me && current_user != presenter.user %>
+    <% if presenter.display_hire_me && !presenter.user_same_as?(current_user) %>
```
tansaku: could the  `!presenter.user_same_as?(current_user)` be pulled into the presenter.display_hire_me method?  no need to expose the logic in the view right?  I guess maybe that could be another ticket and PR but then again ... up to you :-)

diraulo:
@tansaku @Arepo Should I assume all other refactorings will be dealt with as part of a separate PR and merge this?

Arepo:
I'll fix the first one and see if I can do the second as quickly, else I'll pass it to a new PR

tansaku:
@Arepo great work - I think this is ready to go!

tansaku:
@diraulo I think this can be merged - @Arepo did all the refactorings

diraulo:
Thanks for this @Arepo :wink:

PR #1370, opened 2016-10-28
make manually updated hangouts stay live for the duration

### LINKED ISSUE #1358: setting manual hangout link on scrum only keeps event "live" for two minutes
the hangout link updates now, but the event only stays "live" for two minutes - this feels like a resurgence of a previous issue ...

tansaku:
fixes #1358

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ zeus.json
 
 .pryrc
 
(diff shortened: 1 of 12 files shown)


tansaku:
this should also make manual overrides ping slack

PR #1378, opened 2016-11-01
only check for event liveness on day of event

### LINKED ISSUE #1377: when an event is live it shows all future events live as well
same event multiple times

tansaku:
fixes #1377


tansaku:
@diraulo this might also improve performance on the events page ...

PR #1381, opened 2016-11-03
Regenerate friendly ids on projects when title is updated

### LINKED ISSUE #828: Generate friendly project ids
As a project owner
When I rename the project
I would like a new friendly Id to be created for the project
So that I can have consistent "branding" for the project.

Example: Project Unify was originally called Refugee Tech. 
The Project was renamed but the url is still http://www.agileventures.org/projects/refugee_tech

Arepo:
Fixes the bug described at https://github.com/AgileVentures/WebsiteOne/issues/828 

fixes #828

### DIFF
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -1,0 +1,88 @@
+# FriendlyId Global Configuration
+#
+# Use this to set up shared configuration options for your entire application.
+# Any of the configuration options shown here can also be applied to single
+# models by passing arguments to the `friendly_id` class method or defining
+# methods in your model.
+#
+# To learn more, check out the guide:
+#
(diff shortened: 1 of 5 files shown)


config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,88 @@
+# FriendlyId Global Configuration
```
Arepo: It autogenerated this file - shall I leave it as is, or delete the commented cruft?

tansaku: I think maybe leave it - it's got some non-commented cruft that 's maybe required or at least sensible for the latest version of the gem ...

tansaku:
wonderful small PR @Arepo - looks good to go to me :-)

PR #1388, opened 2016-11-10
[SPIKE - DO NOT MERGE] Check to see a video is present before tweeting about it

### LINKED ISSUE #1094: Update tweet service to check if the event is being recorded before posting a tweet with YT link.
```
As a system owner
In order to avoid confusing our Twitter followers
I would like the tweet service to only tweet YouTube links if the event has been recorded
```

Okay, so we have an issue on our Twitter account with users not recording their HOA's and thus, a lot of tweeted YT links are dead. The best scenario is, of course, if all events are being recorded, but that will never be the case so we need to condition this.

Arepo:
fixes #1094

Ok, this is pseudo-finished. That said, it's trying to hit the Google API on a number of specs. Having asked Sam to set up the env variable for the Google API, it occurs to me now that the easiest way to check I've successfully stubbed all the places where it might be doing so is for someone to temporarily remove/distort the API env token and then rebuild on Semaphore so I can cross-reference errors against the specs I've changed.

Does that sound sane, or am I missing a more sensible way of doing it?


app/models/event_instance.rb
```diff
@@ -48,8 +48,26 @@ def broadcaster
...
 
+  def has_video?
+    # The second check is a hack on the gem to prevent having to rescue an error.
+    # I might try opening a PR to the gem itself for this
+    min_plausible_vid_length = 2
+    self.yt_video_id? && video.content_details.any? && video.duration > min_plausible_vid_length
+  end
+
+  def yt_video_id=(video_id)
+    super
```
mattlindsey: Why the super? Necessary?

app/models/event_instance.rb
```diff
@@ -48,8 +48,26 @@ def broadcaster
...
+    self.video = Yt::Video.new id: video_id
+  end
+
+  def video
+    @video ||= Yt::Video.new id: yt_video_id
```
mattlindsey: This is probably fine, but could the 'new' happen in 'initialize'?

mattlindsey:
@tansaku I know we talked about refactoring this, but it actually follows the coding style established by the 'SlackService' class, so I'm recommending we merge this 'as-is.'
Otherwise we should refactor both the Twitter code and Slack code.  Let me know what you think?

mattlindsey:
I just noticed that the cucumber tests still need work.  There are 3 'pending' steps to be implemented.

tansaku:
@mattlindsey yes - we should review this now we have the base tests in place

PR #1392, opened 2016-11-13
Add create event link to navbar

### LINKED ISSUE #822: Creating an event needs to be more obvious
Having to navigate to the events page and then click the little button on the top right is not obvious.  I would include create event in the drop down menu, and also possibly add something somewhere that says like "Let's pair program now" or something.

marianmosley:
Fixes #822


mtc2013:
@tansaku @diraulo ready for feedback

diraulo:
@mtc2013 @marianmosley is the following test passing for you? Keeps failing on CI

```
cucumber -p semaphoreci features/newsletter_form.feature:49
```

mtc2013:
@diraulo i think i got a fix locally.  will push if works out.

mtc2013:
@diraulo fix is in

PR #1418, opened 2016-11-22
add codeclimate config file and linter config

diraulo:
No ticket. Updating configs for codeclimate.


tansaku:
cool @diraulo - didn't realise we didn't already have some of these explicitly - guess we can tweak later as appropriate

tansaku:
Feel free to merge when ready

PR #1422, opened 2016-11-23
updated bootstrap-timepicker to restore icons (#1354)

### LINKED ISSUE #1354: Time-picker's icons are not displayed on events create page (Again 😞)
1. [Create new event](http://www.agileventures.org/events/new)
2. Set a time for the event
3. Note missing icons to increment/decrement

We've had this issue before see #755

c1505:
Fixes #1354


tansaku:
awesome @c1505 - I guess we can't have a test that checks we won't lose them again?

c1505:
@tansaku I am not opposed to having a test for this.  I have never tested for presence of an icon so I wasn't sure if that was a thing to do.  Can you point me in the right direction?

This particular issue was caused because there was some customization added to vendor javascript to fix the problem when it happened before and that was not accounted for when bootstrap-timepicker was moved from vendor to being managed with npm.  It seems the previous time it happened was because an update to font-awesome changed naming conventions.  

This issue does raise 2 general questions for me:
Would we be better off specify version numbers for gems in production?  Are there some better ways of managing and/or eliminating dependencies that we should move towards?

PR #1423, opened 2016-11-23
add delete account functionality for user

### LINKED ISSUE #1421: Add option for users to delete their account
As a user
So that I can not use or be seen on the site
I want to be able to delete/deactivate my account

mattlindsey:
#### Description
This change adds and enables a button on the user's profile to delete their account, which actually does a 'soft delete' due to a previous PR.

#### Motivation and Context
fixes #1421 

Locally I deleted several accounts and saw that they disappeared from the website, but the record still existed with deleted_at set

#### Types of changes

#### Checklist:


features/step_definitions/user_steps.rb
```diff
@@ -402,3 +406,7 @@
   link = user_path(user)
   page.find(:css, %Q{a[href="#{link}?tab=activity"]}).trigger('click')
 end
+
+Then(/^I should not exist as a user$/) do
+  expect(@user).to be_nil
```
tansaku: I think we should reload user from db here to avoid possibility that we are dealing with cached data

tansaku:
@diraulo @mattlindsey tested this locally and it all worked  - made a minor comment, but this is basically ready to go - argument could be made from a UX POV to put deactivate the account from within the edit view since we want to avoid people deactivating accidentally ...

features/step_definitions/user_steps.rb
mattlindsey: hi @tansaku.  I think this is ok.  and I'm not sure how to reload the user given that @user has been destroyed by the delete_user method.

tansaku: @mattlindsey there is a with_deleted method that allows us to reload it - see the README for the gem

tansaku:
@mattlindsey - biggest priority here I would say is moving the functionality into the edit screen - can you do that?

features/step_definitions/user_steps.rb
mtc2013: i think, offhand, it's a scope attached to the model like -- User.with_deleted.find_by(name: "blah") @mattlindsey @tansaku

PR #1439, opened 2016-12-03
add feature to allow user to delete self

### LINKED ISSUE #1421: Add option for users to delete their account

mattlindsey:
#### Description
This change adds and enables a button on the user's profile edit screen to delete their account, which actually does a 'soft delete' due to a previous PR.

#### Motivation and Context
fixes #1421 

Locally I deleted the random account and saw that they disappeared from the website, but the record still existed with deleted_at set

#### Types of changes

#### Checklist:


features/step_definitions/user_steps.rb
```diff
@@ -402,3 +406,8 @@
   link = user_path(user)
   page.find(:css, %Q{a[href="#{link}?tab=activity"]}).trigger('click')
 end
+
+Then(/^I should not exist as a user$/) do
+  user = User.find_by(first_name: 'Anders')
+  expect(user).to be_nil
```
tansaku: can we also check the user exists with deleted_at?  or maybe that's out of scope here?

tansaku:
thanks @mattlindsey for putting this in the edit panel - big improvement - also the test is tighter now - excellent ... probably ready to pull in ...

tansaku:
I'm checking on the automated deploy and seeing:



Which looks great!

features/step_definitions/user_steps.rb
tansaku: I guess the test would still pass if we had the more thorough destroy - would be good to check we can definitely undo ...

tansaku: let's put that into a refactoring ticket https://github.com/AgileVentures/WebsiteOne/issues/1455

PR #1441, opened 2016-12-05
1402 refactor charges to subscriptions controller

### LINKED ISSUE #1402: refactoring - charges controllers should really be subscriptions controller
what is currently called the charges controller, actually manages the creation of subscriptions ... should change for clarity and move out things like updating card details into a cards controller (or something ...)

mtc2013:
fixes #1402


tansaku:
we just noticed we missed one thing: 

https://github.com/AgileVentures/WebsiteOne/blob/a921361232c002ee89e880e0659789e5a9a202c4/app/controllers/subscriptions_controller.rb#L37

which also highlights that we probably don't have a sadpath test for this ...

tansaku:
@diraulo - this is ready for review - that CodeClimate issue is just mis-interpreting our file I believe

PR #1459, opened 2016-12-14
Have paypal payment update our server after completed payment

### LINKED ISSUE #1438: paypal payments don't update our model on the server
need to connect paypal payment hook to come through on our server to maintain correct back end representation

tansaku:
fixes #1438


app/helpers/subscriptions_helper.rb
```diff
@@ -1,2 +1,15 @@
 module SubscriptionsHelper
+  def paypal_form_url
+    return 'https://www.paypal.com/cgi-bin/webscr' if ENV['RAILS_ENV'] == 'production'
```
tansaku: would make more sense in config

PR #1465, opened 2016-12-19
support generic subscription plans

### LINKED ISSUE #1452: add support for generic subscription plans
new plans under development - could we support new plans without having to deploy code?

As an admin
So that I can adapt the charity business fast
I would like to be able to quickly release new plans without deploying new code

mtc2013:
fixes #1452


app/controllers/subscriptions_controller.rb
```diff
@@ -7,7 +7,8 @@ class SubscriptionsController < ApplicationController
   def new
     @upgrade_user = params[:user_id]
     @sponsorship = @upgrade_user && current_user.try(:id) != @upgrade_user
-    render plan_name
+    plan = params[:plan] || 'premium'
```
mattlindsey: Rather than hardcoding your 'default' plan, you could indicate which plan is default by a flag in the new Plans table.  Then you could change the terms of the default plan for new people while keeping existing people on the plan they came in on.

tansaku: interesting idea @mattlindsey - thanks!

db/schema.rb
```diff
@@ -11,20 +11,21 @@
...
+ActiveRecord::Schema.define(version: 20161221125828) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pg_stat_statements"
 
   create_table "activities", force: :cascade do |t|
     t.integer  "trackable_id"
-    t.string   "trackable_type"
+    t.string   "trackable_type", limit: 255
```
tansaku: I expect it's because I upgraded my local PGsql

lib/tasks/create_plans.rake
```diff
@@ -0,0 +1,8 @@
+desc "This task creates the set of premium plans"
+
+task :create_plans => :environment do
```
mattlindsey: namespace 'db' here?

mattlindsey:
This PR looks good to me, although a person can only 'sponsor' another for premium -- not mob or f2f.  Right?

tansaku:
@mattlindsey yeah, at the moment it's only sponsor others for premium, but would like to make it sponsor at any level, but first we need to fix it so people can upgrade smoothly through the levels - we're still only allowing upgrade to premium plus through out web interface :-(

lib/tasks/create_plans.rake
tansaku: good point

johnnymo87:
Some small nitpicks, maybe for a future PR:
* In the controller, what happens if `find_by` returns `nil`?
* In the controller, the params handling (e.g. `|| 'premium`, `.downcase.delete(' ')`) suggest that there's an object hidden here.
* Related to the two previous bullets, perhaps a `Finder` object could handle the params parsing and the query lookup. If it encounters problems, it throws domain-specific errors, which your controller `rescue_from`s and forwards to the user. Otherwise, it provides a clean interface for "params in, entity out".
* The mail and view templates have too much knowledge. Ideally, they'd be handed a presenter with all the info they need one method call away. Tip: `number_to_currency` can be found outside a template at `ActiveSupport::NumberHelper.number_to_currency`.
* Is there any reason not to have a null constraint on every single column you're adding?
* Maybe I missed it, but is there any validation on what `third_party_identifier` must be? For increased type safety, you could could use an enum defined by activerecord and postgres: https://www.sitepoint.com/enumerated-types-with-activerecord-and-postgresql/

tansaku:
@johnnymo87 wrote:

> In the controller, what happens if find_by returns nil?

we get an error --> made a ticket: https://github.com/AgileVentures/WebsiteOne/issues/1488

> In the controller, the params handling (e.g. || 'premium, .downcase.delete(' ')) suggest that there's an object hidden here.

The papal item_name is displayed in the PayPal checkout page e.g. "Premium Mob" - we also have to use that as the "id" - I think there's room here to extract PayPalProcessCompleteService and StripeProcessCompleteService entities here 

> Related to the two previous bullets, perhaps a Finder object could handle the params parsing and the query lookup. If it encounters problems, it throws domain-specific errors, which your controller rescue_froms and forwards to the user. Otherwise, it provides a clean interface for "params in, entity out".

agreed - I'm thinking of a subsequent services refactoring

> The mail and view templates have too much knowledge. Ideally, they'd be handed a presenter with all the info they need one method call away. Tip: number_to_currency can be found outside a template at ActiveSupport::NumberHelper.number_to_currency.

good point

> Is there any reason not to have a null constraint on every single column you're adding?

Not sure what you mean - i see a null constraint on timestamps - I think just generated as part of the migration.  Haven't explicitly gone through adding null constraints.  Perhaps I'm missing what you are referring to?

> Maybe I missed it, but is there any validation on what third_party_identifier must be? For increased type safety, you could could use an enum defined by activerecord and postgres: https://www.sitepoint.com/enumerated-types-with-activerecord-and-postgresql/

Currently no validation - the third_party_identifier is just a downcased, underscored, computer-friendly version of the human-friendly version of the plan name.  The idea is that we can support new plans (with new names and ids) without having to make any code changes, i.e. simply adding a row to the database should be sufficient to support subscription to a new plan

johnnymo87:
> > Is there any reason not to have a null constraint on every single column you're adding?

> Not sure what you mean - i see a null constraint on timestamps - I think just generated as part of the migration. Haven't explicitly gone through adding null constraints. Perhaps I'm missing what you are referring to?

I mean to say, why not have a null constraint on everything, instead of just timestamps?

tansaku:
@johnnymo87 I guess we could have the not null constraint in many places ... although I'm a little wary about adding validations and data-integrity stuff when it's not driven from the user.  The set of validations we tried to drive through for organisation urls, emails etc. was difficult to complete and I'm not it delivered any value to the client ...

PR #1474, opened 2016-12-29
Display upcoming events during event duration (fixes #1467)

### LINKED ISSUE #1467: live events on events page should stay while they are live
this could be either from their duration, or as long as they stay live

c1505:
Fixes #1467 
Comments and changes welcome.  

Events were not displayed 15 minutes after the start of the event regardless of the event duration or if the event is associated with a live EventInstance(google hangout). This was changed to display events throughout the duration of the event and beyond the duration of the event if the event is still live.  I also moved some existing logic from the controller to the model.


tansaku:
great work here @c1505 - I'm not sure what happened to the CodeClimate thing - I've emailed their support to ask why it's not building there - the output doesn't make any sense ...

app/models/event.rb
```diff
@@ -18,7 +18,7 @@ class Event < ActiveRecord::Base
   attr_accessor :repeat_ends_string
 
   COLLECTION_TIME_FUTURE = 10.days
-  COLLECTION_TIME_PAST = 15.minutes
+  COLLECTION_TIME_PAST = 300.minutes
```
tansaku: if this is the fix that's awesome - I guess what this means is that any event that was started in the last 5 hours will stay on the upcoming events page.    Ah, but you're handling that with the `remove_past_events` method - I see

tansaku:
great work @c1505 - this looks like it could be a great solution.  I'm interested to see what the performance is like on our production data.

Also, should we have some cucumber tests here to really explain the different circumstances, e.g. scenarios like:

* an event that's still live after 3 hours
* an event that's still within duration
* an event that's out of duration
* etc.

You've got some new boolean logic that I had to look through to infer what would happen, but that's where I think cucumber is great for making really explicit.

Are you familiar enough with Cucumber to take that on by yourself or would you like some support?

tansaku:
@c1505 Code climate are asking if we can push a new commit to this branch as they've never seen this error before :-) I can do it myself if you give me permissions on your repo ...

c1505:
Thanks @tansaku .  You are a collaborator on my fork.  Are there other permissions besides that?  

As far as tests go, I was thinking cucumber first, but added model specs instead because I thought it would be faster and still provide sufficient coverage.  As long as we are getting the appropriate events with the upcoming_events method, won't that cover our bases?  There are already some cucumber tests for upcoming events in list_events.feature .  Either way, I should go back and make sure the edge cases are covered better and potentially re-name tests for clarity.

c1505:
Well fixed one broke one :) .
Looks like at least one of these has failed intermittently on recent builds from other PRs.  edit_hangout_url.feature:18 Failed once for me when I ran it locally and is now passing again.  Has anyone looked into either of these?  

Failing Scenarios:
cucumber -p semaphoreci features/hangouts/edit_hangout_url.feature:18 # Scenario: Edit Hangout URL and ensure event stays live
cucumber -p semaphoreci features/projects/edit_project.feature:58 # Scenario: Editing Pitch content with Mercury Editor

c1505:
Thanks @tansaku for all the explanation.  I am here to learn :) Still trying to figure out what and how to test in general plus I understand different groups will have different testing strategies.  

My understanding now is that the user storie(s) we are implementing here are: 

As a site user
  So that I can find events to join
  I want to see upcoming events
  And I want to see live events
  And I want to see events that are within the scheduled event duration
  
Should I also be testing with cucumber that I don't see these events when they don't meet that criteria or would that be better done in the specs?

tansaku:
@c1505 - I think that they key is to have those three elements you mention in the cukes - at the moment we have only one.  Then there's the question of how many "sad path" elements we want to put in the cukes.  We don't want too many as the cukes are slow running, but it seems natural to me to have the following scenarios:

Scenario: List still shows an event that's live after 3 hours when duration was 1 hour
Scenario: List still shows an event that's within it's duration of 1 hour, 10 minutes before end
Scenario: List does not show an event that's out of duration
Scenario: List does show upcoming events that haven't started

I've maybe been too specific with some of the wording there, but I'd say that they pretty much cover the key intention we have for the events list for these changes, right?

Also, it would be fine to break out a separate refactoring ticket to try and make the existing single scenario that covers "List does show upcoming events that haven't started" more declarative - if that makes sense ... maybe I should do that to help?

Another thing is to treat your initial work as a spike, and re-drive it's creation from acceptance tests in a new branch - we've done that on some other features.  Again it's all about how much you want to take on.  Let me know if you want more input on things like imperative vs declarative etc.

It's also important we don't load you with too much stuff :-) and work out the way to slice to work to allow you to close out the work in stages ...

c1505:
I would like to work on the cucumber tests too.  I'll update the PR as I go along, but no need to review again yet.  I'll comment again when I think it is ready to be reviewed.

c1505:
Ok.  I think this is ready for another look now.  A bit unsure about the new cucumber tests and also welcome to other changes too.  @tansaku

tansaku:
Looks just like what the doctor ordered @c1505 - I'll take a more detailed look too :-)

tansaku:
@c1505 a closer look also looks good.  Great that you've structured this all using existing cucumber steps.  The thing to consider is whether any of the cucumber steps could pass in the absence of any of the functionality?  Have we been able to see these cukes fail so that we can sure that they will properly fail when something breaks.

An interesting exercise might be to try introducing the cuke into a new branch forked from develop and running the cukes there (check they fail) then add the specs (check they fail) and then add the functionality to see everything go green ...

c1505:
Looks like I forgot to do that check to make sure they fail without the new code.  Thanks for the reminder.  Here is the branch with just the new fixed cucumber tests https://github.com/c1505/WebsiteOne/tree/just_tests_1474_on_develop .    Held off on squashing commits in case there are additional changes to be made.

c1505:
Anything I should add or change here?

tansaku:
@c1505 - you said something about the updated cukes being in a different branch - maybe we need a new pull request? to get that new code in ...?

c1505:
I updated them in this branch.  The other branch only contains the updated cukes to show that they fail without the new code.  I ended up squashing my commits on this branch because I thought it was ready to go.  Should I be doing that or is it preferred that I leave the PR with multiple commits?

tansaku:
@c1505 in general I'd leave them with multiple commits so we can see the process :-) also, it should save you the effort since GitHub can now squash them for you - great work getting the other cukes in a separate branch - would be good to automate that somehow for all tests ....

tansaku:
Would be nice if I could quickly get to a CI build of that other branch - can't seem to see one - we have it set up on LS for that - I guess that's a benefit of travis, or just a setting we have there ...

mattlindsey:
Code looks good to me :).
I tried to test on my local machine.  When I created a new event and started the hangout, it didn't show the event as live in the upper left of the screen as expected.  But maybe I wasn't testing properly?

mattlindsey:
Looks good to me.

c1505:
Thanks @mattlindsey  :)

Issue #1475, opened 2016-12-29
Automatically create an event when an EventInstance(google hangout) is created 


c1505:
## Current Behavior
A lot of google hangouts are started without having a related event.  Because of that, they aren't visible on the upcoming events page.    

## Possible Solution
Create an Event every time an EventInstance is created and associate that Event with the EventInstance.

c1505:
As a user
  So that I can join live events
  I want to see live events on the upcoming events page even when they are not started from an event

c1505:
This is on pause for me until more is known about the future of the google hangout api.  Another person is welcome to work on it if they would like.

PR #1478, opened 2017-01-04
Experimental spike auto create events with EventInstance creation issue #1475

c1505:
Issue #1475 .  Work in progress.  I want to see if I should be going about this in a different way.  

As a user
  So that I can join live events
  I want to see live events on the upcoming events page even when they are not started from an event

For my first stab at it, I am just using an ActiveRecord callback to create an associated event if one does not exist.  I have heard it is bad practice to use ActiveRecord callbacks for things like this, but I have not felt the pain of doing so.  Could use some advice.


app/controllers/event_instances_controller.rb
```diff
@@ -3,14 +3,18 @@ class EventInstancesController < ApplicationController
...
+    event_instance = EventInstance.find_by(uid: params[:id])
+    # another option could be to just call a different method on the instance.
 
     event_instance_params = check_and_transform_params(event_instance)
     hangout_url_changed = event_instance.hangout_url != event_instance_params[:hangout_url]
     yt_video_id_changed = event_instance.yt_video_id != event_instance_params[:yt_video_id]
     slack_notify = params[:notify] == 'true'
 
-    if event_instance.try!(:update, event_instance_params)
+    if event_instance.try!(:update, event_instance_params) #why would it not update.  is it only if things changed?
```
tansaku: I think this is caution because the data is coming in from a google hangout plugin and we're trying to avoid a failure on the off chance that event_instance would be nil - but that said this is really gnarly code -- there's another PR coming through that cleans up a little here so I'd be cautious about working on this  - I might suggest that it would be more effective to review the existing open PRs and get your head around what other team members are doing than trying to make code changes here.

That said, don't let me stop you playing with experimental PRs in any part of the codebase, just that in some areas where existing work is going on the changes of challenging merge conflicts is much higher than others ...

app/models/event_instance.rb
```diff
@@ -1,4 +1,5 @@
 class EventInstance < ActiveRecord::Base
+  after_create :create_related_event
```
tansaku: the pain of these callbacks usually comes when they are triggered unexpectedly - I've had enough of that pain to want to avoid them :-)

I'd prefer to have a service object codify the business logic relationships between models and have that service object be called from the controller

tansaku:
thanks for taking a stab at this @c1505 

one general rule of thumb (that I often end up breaking) is to try and avoid working on two things at the same time on the same project.  one might argue that it would be good to finish up your existing PR before starting on this one, but if you're really keen to spike something and are happy that they code might get thrown out then that's all good :-)

The danger with multiple PRs is that the longer they stay unmerged, the greater the chances of getting more are more difficult merge conflicts as the rest of the codebase changes.

A couple of other things I notice here looking over your code (besides the comments I added inline) are that you're doing this PR off of the code from your other PR branch, which I think mixes things up.  If you're starting to look at a new thing it's probably best to branch off of develop to make it easier to see which changes are new to this new PR.  That's also another reason to want to get an outstanding PR fixed and merged first so that that code is already in develop for a new clean branch.

Also, while you've included a great user story in this PR, you're not driving from acceptance tests.  While that's fine for spiking purposes, there's an argument that we want to describe the end behaviour of the system in terms that are not specific to the implementation.  There's sometimes a danger of thinking about the implementation too soon in that it prejudices us about how we will implement something, when the key goal is not the nature of the implementation per se, but the end behaviour.  Sometimes if we come at the problem from the angle of the end behaviour that the user experiences then we'll leave ourself open to new insights about how to implement that behaviour ... 

Just a few thoughts :-)

tansaku:
@c1505 all good - maybe we need a spike label :-)

mattlindsey:
hi @c1505.  Are you looking for more comments on this spike?  If not, maybe it can be closed?

c1505:
I am ok closing this.  I would rather wait to try and implement this feature with the uncertainty that there is around the google hangouts api and moving events to a microservice.  I still think it is a desirable feature, but seems likely that the implementation would need to change a few months from now.

PR #1480, opened 2017-01-05
Fix syntax error in schema.rb

ragesoss:
With the stray comma, the schema file does not load.


tansaku:
@ragesoss thanks for the PR - I just merged a conflict - assume this will be okay

not clear about how this sneaks past our CI - I guess CI does migrate, but maybe @ragesoss did something else locally ...?

PR #1492, opened 2017-01-12
Change logic to depend on actual_start_time

### LINKED ISSUE #1489: should be checking actual start time of livestream and not duration in order to tweet streams
duration stays 0 till stream finishes.

We should prefer `live_streaming_detail.actual_start_time` which is nil before stream starts:

```
irb(main):024:0> unstarted = Yt::Video.new id: 'iLFHzaCplK0'
=> #<Yt::Models::Video:0x007f82349343c0 @id="iLFHzaCplK0", @auth=nil>
irb(main):025:0> unstarted.live_streaming_detail.actual_start_time
```

mtc2013:
Fixes #1489

### DIFF
--- a/app/services/twitter_service.rb
+++ b/app/services/twitter_service.rb
@@ -95,7 +95,7 @@ module TwitterService
     uri = URI.parse("http://gdata.youtube.com/feeds/api/videos/#{code}")
     Net::HTTP.get(uri)
     video = Yt::Video.new id: code
-    return true if video && video.duration > 2
+    return true if video.live_streaming_details.first.try(:actual_start_time)
     return false
   end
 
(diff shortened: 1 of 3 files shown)


mtc2013:
@tansaku it seems like payment tests unrelated to this PR have broken.

tansaku:
@mtc2013 failure looks like it came from server trying to reach:  https://www.sandbox.paypal.com/en_GB/i/scr/pixel.gif  - puffing billy required perhaps ...?

app/services/twitter_service.rb
```diff
@@ -95,7 +95,7 @@ def self.valid_recording(code)
     uri = URI.parse("http://gdata.youtube.com/feeds/api/videos/#{code}")
     Net::HTTP.get(uri)
     video = Yt::Video.new id: code
-    return true if video && video.duration > 2
+    return true if video.live_streaming_details.first.try(:actual_start_time)
```
tansaku: I note I was using `unstarted.live_streaming_detail.actual_start_time` to check the actual start time in the console ...

PR #1500, opened 2017-01-15
Adding new language for organisation plans

### LINKED ISSUE #1499: add different language for organization plans
signing up for an organisation plan needs slightly different language than that for an individual

tansaku:
fixes #1499


tansaku:
initial round trip - we also need to look at:

* [ ] rest of language in template
* [ ] language in email
* [ ] think about level of logic in view --> move to presenter?  move strings to yml file?

db/schema.rb
```diff
@@ -248,44 +249,44 @@
...
     t.datetime "updated_at"
-    t.string   "first_name",             limit: 255
-    t.string   "last_name",              limit: 255
+    t.string   "first_name"
+    t.string   "last_name"
     t.boolean  "display_email"
-    t.string   "slug",                   limit: 255
-    t.string   "youtube_id",             limit: 255
-    t.boolean  "display_profile",                    default: true
+    t.string   "youtube_id"
```
mattlindsey: Hi @marianmosley.  Did you mean to reverse 'youtube_id' and 'slug'?

db/migrate/20161221182758_remove_stripe_customer_from_user.rb
```diff
@@ -1,5 +1,5 @@
 class RemoveStripeCustomerFromUser < ActiveRecord::Migration
   def change
-    remove_column :users, :stripe_customer
+    remove_column :users, :stripe_customer, :string
```
mattlindsey: Is the last parameter a mistake (":string")?

tansaku: no - it's needed to allow the migration to be reversible

db/schema.rb
tansaku: why do you think it's reversed @mattlindsey ?

tansaku:
before we merge this we really should:

* [ ] review all the rest of the text in the interface
* [ ] add tests to check the emails contain the right text

PR #1510, opened 2017-01-20
Fix last_github_update in projects

### LINKED ISSUE #1506: Wrong date displayed for project Last Update
Line should be:
last_github_update: client.repo(github_url(project)).pushed_at)

mattlindsey:
fixes #1506


app/jobs/github_commits_job.rb
```diff
@@ -8,6 +8,7 @@ def run
       begin
         update_total_commit_count_for(project)
         update_user_commit_counts_for(project)
+        update_last_pushed_dt_for(project)
```
tansaku: so perhaps we want to have these in separate rescues so one failure doesn't block the others?

app/jobs/github_commits_job.rb
```diff
@@ -8,6 +8,7 @@ def run
...
       rescue Exception
```
tansaku: And we might want to replace this Exception with StandardError: https://robots.thoughtbot.com/rescue-standarderror-not-exception I know you didn't introduce this, but a great time to fix it perhaps ...

PR #1525, opened 2017-01-24
First brush at call to action improvement

mtc2013:
Here's a picture of the screen on desktop: 

screen shot 2017-01-24 at 12 19 45 pm


tansaku:
@mtc2013 cool - we could even drop the "let's code"

or go with "Sign up to Starting Coding!" and perhaps it needs to look more like an actual link or button to emphasize it can be clicked?

tansaku:
@mtc2013 from discussions today I wonder if getting people to add their email address for newsletter is even more important than getting them signed up?

tansaku:
but simplest thing first I guess - simplest to add some "sign up" call to action

mtc2013:
@tansaku updated with a better style and test:

screen shot 2017-01-25 at 12 21 56 pm

tansaku:
great - what we were just talking about - I would be tempted to make text larger and reduce width of button - happy to do tweak myself if you'd like to move on to something else

mtc2013:
@tansaku give it a go.  had a bit of trouble getting the text bigger.  I used a bootstrap class to style, and one of the things is it will expand to the containing width (block level) to get that larger size

mtc2013:
Made text larger.  
screen shot 2017-01-25 at 1 15 40 pm

mtc2013:
@tansaku tada

screen shot 2017-01-25 at 1 22 22 pm

tansaku:
liking it @mtc2013 - tempted to start tweaking the color.  Is that our standard blue?  Should we make it lighter? make it change on mouse over?  Don't worry if you've had enough here.

Perhaps we can show to folks in scrums tomorrow and get some input from folks like @lollypop27

mtc2013:
@tansaku perhaps we should push it out first? I mean it seems to be a clear win over what we currently have.  Subtle tweaks can be discussed in scrum and people can look at it on production?

tansaku:
one reason to avoid too many tweaks on production is just for consistency of experience, but yes it could just be pushed.  It will take a little while to move it along the chain anyway, during which people can look at it on develop and staging

mtc2013:
@tansaku on hover now the background of the button becomes that of the underlying containing div.. that kind off offwhite, and the text goes to the buttons former blue on hover.  can't really get a good screenshot of it but if you pull it down and run it you'll see.

tansaku:
thanks @mtc2013 - doh, and I forgot to show in scrum - will get it onto develop to share or something ...

tansaku:
great work on this @mtc2013 - I've got it running locally - one tiny thing was that I found that completely inverting on rollover was actually kind of painful, i.e the sudden white.  I just changed to:



with the roll over going to our deeper blue that's part of the nav bar:



but you're right we should get something out there, and then keep tweaking the color scheme ...

tansaku:
that's with: 

```css
.call-to-action:hover {
    background: #34495e;
}
```

mtc2013:
@tansaku unfortunately this doesn't play well on the mobile browsers

mtc2013:
@tansaku updated with tweaks for various mobile browsers!

tansaku:
@mtc2013 looks good in portrait:



landscape still has an issue:

mtc2013:
@tansaku updated again

PR #1529, opened 2017-01-25
moves card operations out of subscriptions controller

### LINKED ISSUE #1526: refactor card update out of subscriptions controller
it should be in a CardController

tansaku:
fixes #1526


app/controllers/cards_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  private
+
+  def stripe_token(params)
+    Rails.env.test? ? generate_test_token : params[:stripeToken]
+  end
+
+  def generate_test_token
+    StripeMock.create_test_helper.generate_card_token
```
arreche: Is it ok using a mock here in the production code? I expected to see mocks injected in test code only 😄

tansaku: it's not good @arreche but it's necessary for the StripeRubyMock gem https://github.com/rebelidealist/stripe-ruby-mock - I was working without it for a long time using pure VCR/PuffingBilly caching, but ultimately it got really sticky, and stripe ruby mock has been smooth sailing since, so I'm calling it a necessary evil for now ...

tansaku: I wrote several blogs on it if you want to hear the gory details:

https://www.google.co.uk/search?q=site:nonprofits.agileventures.org+stripe

PR #1533, opened 2017-01-26
makes routes for card and subscription restful

### LINKED ISSUE #1532: refactor subscription upgrade to more appropriately named update
this will make more sense in terms of REST and remove a special separate route

tansaku:
fixes #1532


app/controllers/subscriptions_controller.rb
```diff
@@ -35,6 +23,18 @@ def create
     redirect_to new_subscription_path(plan: (@plan.try(:third_party_identifier) || 'premium'))
   end
 
+  def update
+    customer = Stripe::Customer.retrieve(current_user.stripe_customer_id)
+    subscription = customer.subscriptions.retrieve(customer.subscriptions.first.id)
+    subscription.plan = "premiummob"
```
arreche: Types and symbols are less error prone that hardcoded strings ;)

Please. Consider refactoring those `premium` and `premiummob` plan types too.

config/routes.rb
```diff
@@ -6,9 +6,7 @@
   resources :activities
   resources :newsletters
 
-  match '/cards/update' => 'cards#update', :via => [:put]
-  match '/cards/:user_slug/edit' => 'cards#edit', :via => [:get]
-  match '/subscriptions/upgrade' => 'subscriptions#upgrade', :via => [:put]
```
arreche: So. What is the `subscriptions` resource now? Does it have a more appropriate name after the refactor?

tansaku: it's subscriptions/update - it's folded into resources :subscriptions - properly RESTful - well maybe - what the user is doing is more like account or user update - they are changing something about the nature of their relation with AV ...

app/controllers/subscriptions_controller.rb
tansaku: Yeah, that's being pulled out in the next step for generic updates - however that string is the 3rd party stripe id and that subscription object is from the Stripe API so we can't quite change it - we can only wrap it

PR #1542, opened 2017-01-31
1536 hide call to action when signed in

### LINKED ISSUE #1536: Should we make the button not display when the user is signed in already?
screen shot 2017-01-30 at 12 13 43 pm

mtc2013:
Fixes #1536
screen shot 2017-01-31 at 12 29 06 pm

### DIFF
--- a/app/views/layouts/_round_banners.html.erb
+++ b/app/views/layouts/_round_banners.html.erb
@@ -1,7 +1,9 @@
 <!-- BEGIN LIST -->
 <ul class="box">
+  <% unless current_user %>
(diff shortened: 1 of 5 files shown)


app/views/layouts/_round_banners.html.erb
```diff
@@ -1,7 +1,9 @@
 <!-- BEGIN LIST -->
 <ul class="box">
+  <% unless current_user %>
```
tansaku: should probably lose prepended whitespace here

features/step_definitions/user_steps.rb
```diff
@@ -119,6 +119,7 @@
 ### THEN ###
 Then /^I should be signed in$/ do
```
tansaku: could this step name tell us more accurately about what it is doing under the hood?

mtc2013:
@tansaku updated to address concerns

PR #1545, opened 2017-02-02
throws adwords script in premium created view

### LINKED ISSUE #1544: get adwords conversion tracking in
as this will allow us to see which of our adwords campaign are leading to conversions

This includes
* [ ] member sign up
* [ ] premium sign up
* [ ] premium upgrades

tansaku:
fixes #1544


tansaku:
I've checked this appears - unclear if it will have the effect desired in it's current location - will have to wait and see ...

tansaku:
maybe this should only be shown in production?  hopefully it has to be served from av.org to be added to adwords count ...

PR #1551, opened 2017-02-07
adjust mail to link in user summary

### LINKED ISSUE #1495: The email link on a members pages results in a 404 error
On a member's page where there is an email link, e.g.
http://www.agileventures.org/users/patrick-bolger

When I click on the email link (http://www.agileventures.org/users/patmbolger@gmail.com) I get a 404 error.

I would expect this link to either cause my email program to open up to send them an email, or that an email form is opened up in the website for me to send them a message.

tansaku:
fixes #1495


tansaku:
so @ManuelCostanzo - if we merged this we'd get the change we wanted, although we might also want a test as well to automatically catch this issue in future ...

PR #1553, opened 2017-02-08
spikes a flash setting to add the adwords conversion tracking after sign up

### LINKED ISSUE #1549: adword tracking of signups
if we want to track adwords leading to signups we need to put some google script into the page the user lands on after sign up.

I think that's the home page at the moment - maybe we should redirect them to "Getting Started" and insert that script depending on having arrived from the sign up process?

Devise has hooks to redirect:

https://github.com/plataformatec/devise/wiki/How-To:-redirect-to-a-specific-page-on-successful-sign_in,-sign_out,-and-or-sign_up

tansaku:
fixes #1549


arreche:
Elegant solution 👍

PR #1564, opened 2017-02-27
Tooltips to explain Premium upgrade buttons

### LINKED ISSUE #1530: add tooltips for "upgrade to mob" to explain and/or add info buttons
```
As an AgileVentures Admin
So that more people will sign up for premium and upgrade to premium mob
I would like the benefits of upgrading to be easily discoverable from the individual user pages
```

these buttons are appearing on user profiles - would be great to make their meaning discoverable

tansaku:
fixes #1530

### DIFF
--- a/app/views/users/profile/_premium_mob_upgrade.html.erb
+++ b/app/views/users/profile/_premium_mob_upgrade.html.erb
@@ -2,7 +2,9 @@
     <% button_text = 'Upgrade to Premium Mob' %>
     <%= form_tag subscription_path, method: "put" do %>
         <input type="hidden" name="user_id" value="<%=presenter.user.slug %>">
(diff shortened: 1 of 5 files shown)


tansaku:
not sure if this even warrants a test

certainly could move this long text into l18n locales ...

tansaku:
I remember doing it one time in LocalSupport - here's a relevant SO: http://stackoverflow.com/questions/17880922/testing-a-html-class-title-tag-with-capybara

tansaku:
in LocalSupport we had:

```
Then /the following tooltips should exist/ do |tooltips_table|
  tooltips_table.hashes.each do |tooltip|
    expect(page).to have_xpath("//tr/td[@title='#{tooltip['tooltip']}'][@data-toggle=\"tooltip\"]")
  end
end
```

PR #1566, opened 2017-02-28
Less intimidating invitation to scrums for non-logged in users

### LINKED ISSUE #1554: adjusting how "next scrum" is displayed to new users
```
As an AgileVentures Admin,
So that new members are more likely to watch and/or attend scrums and other events
I would like the language in the upcoming event notification to me as welcoming as possible, i.e. not mention things like "scrum" and "kent beck" that may be unfamiliar
Whereas long term members should still see the original format???
```

```
As a new member
So that I can understand better how the community works
I want clear instructions about how to get help and support
```

as we were discussing in the [scrum](https://www.youtube.com/watch?v=eUhIV1c9CpU) we'd like to adjust the messaging on the upcoming scrum for those who are not logged in ...

Currently it says things like:

> "Kent Beck" Scrum & Pair Hookup (Scrum Master: Sam Joseph) in 2 hours 3 minutes

we'd like to change it to something like:

> “Want to learn more?  Listen in during the next projects review meeting, in 2 hours 3 minutes.”

when events are live there is other text, which we might also change to make it more inviting and less intimidating.  In the first instance we're just planning to do this for users that are not logged in.

tansaku:
fixes #1554


app/helpers/layout_helper.rb
```diff
@@ -12,5 +12,9 @@ def flash_messages(opts={})
   def show_layout_flash?
     @layout_flash.nil? ? true : @layout_flash
   end
+
+  def event_name_or_invitation_to_guest_user(event)
+    current_user ? event['name'] : 'Want to learn more? Listen in. Next projects review meeting '
```
tansaku: I wonder if this string should be moved into our `config/locales/en.yml` and then we refer to that constant in the cucumber steps rather than the specific output

app/views/layouts/_navbar.html.erb
```diff
@@ -123,7 +123,7 @@
   <% if @next_event %>
     <div class="hidden-sm hidden-xs hidden-md" id="next-event" data-event-time="<%= @next_event.next_occurrence_time_attr.to_datetime %>"
          data-event-url="<%= event_path(@next_event) %>"
-         data-event-name="<%= @next_event.name %>"></div>
+         data-event-name="<%= current_user ? @next_event.name : 'Next projects review meeting - listen in to learn more about AgileVentures' %>"></div>
```
tansaku: this could also be replaced by a helper, but text needs to be slightly different for this countdown ...

features/events/next_scrum_for_new_user.feature
```diff
@@ -0,0 +1,50 @@
+@vcr
```
tansaku: I'm concerned this file is a heavyweight replication of `next_scrum.feature` which just checks a different text when a user is not logged in - would a view spec be more sensible or sufficient ...?

or a scenario outline?  Or just some tweaking on whether all the set up is needed in the cucumber background?

spec/views/visitors/index.html.erb_spec.rb
```diff
@@ -45,6 +45,7 @@ def fix_time_at(time)
     end
   end
 
+  let(:invite_text) { 'Want to learn more? Listen in. Next projects review meeting' }
```
tansaku: if we were using l18n we could be looking up this text from locale - certainly for a spec that seems appropriate

features/events/next_scrum_for_new_user.feature
tansaku: see notes here: https://github.com/AgileVentures/WebsiteOne/issues/1571

PR #1595, opened 2017-03-14
Add blog link to footer

### LINKED ISSUE #1575: add a link to the nonprofits site
```
As an AV owner
So that we can get more comments on the static site, and get AV members excited, and help them find the blog ...
I would like AV members to be able to navigate to the static site
Unlike now when it is not linked
```

@sampritipanda said: 

> I think there should be a link for http://nonprofits.agileventures.org/ on the main site, I had a hard time finding out where the AV blog was

mtc2013:
fixes #1575 

Should also add to editable about-us page, as suggested.


mtc2013:
@tansaku added aa link to about-us page.

PR #1602, opened 2017-03-16
increase base font size to 24

### LINKED ISSUE #1578: too much text on static pages (font size too small)
```
As a site owner
So that members are not put off my excessive text
I would like the font sizes to be larger for basic text to encourage conciseness
Unlike now where the small body font size encourages long text
```

I wonder if one of the reasons that we have too much text on many of our static pages is that the font size is rather small?

A lot of them look like essays, when I think we might benefit from drifting towards pithier shorter content?

I'm thinking about:

http://www.agileventures.org/about-us
http://www.agileventures.org/code
http://www.agileventures.org/grow
http://www.agileventures.org/learn

and so on ...

would be nice to experiment with a larger text size and use that to prompt reducing the amount of text ...

could also increase the line height

tansaku:
fix #1578

### DIFF
--- a/app/assets/stylesheets/global/custom.scss
+++ b/app/assets/stylesheets/global/custom.scss
@@ -32,7 +32,7 @@ html, body {
   /* The html and body elements cannot have any padding or margin. */
   -webkit-font-smoothing: antialiased;
   font-family: 'Ubuntu', sans-serif;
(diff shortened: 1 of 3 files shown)


tansaku:
hmm, this screws up the home page:



maybe just change this on static controller style?

PR #1612, opened 2017-03-21
move challenge key to env var

### LINKED ISSUE #1611: Renew all SSL certificates
as it will change when we need to renew

tansaku:
fixes #1611


tansaku:
this could do with a test, since we're going to be relying on it in a year to update our SSL keys

@arreche - you were saying something about being able to set up a cron job for that ...?

arreche:
> Let’s Encrypt CA issues short-lived certificates (90 days). Make sure you renew the certificates at least once in 3 months.

http://letsencrypt.readthedocs.io/en/latest/using.html#renewing-certificates

tansaku:
thanks @arreche - once every three months then ... I've opened a ticket #1613

tansaku:
gah, but how does this interaction with https://devcenter.heroku.com/articles/automated-certificate-management ?

PR #1634, opened 2017-04-23
Fix project scope to work for nil project_url and other improvements

### LINKED ISSUE #1632: some users have no github activity tab
No github activity tab, and karma doesn't seem to be increasing.

E.g.,
https://www.agileventures.org/users/matt-lindsey?tab=activity

mattlindsey:
fixes #1632 

Problem is GithubCommitsJob.run is using a db scope that compares projects.github_url to blank(''), which gives active record error.  Checking for nil avoids the error and eliminates most log messages.

After this fix is applied and the job is run, one should be able to verify the fix by going to the following, which should now show activity:
https://www.agileventures.org/users/matt-lindsey?tab=activity


mattlindsey:
Actually it appears that this is only a partial fix, since I should have more commits in other repos:

image

PurityControl:
More of a question than a criticism but given the special treatment that nulls require, what are peoples thoughts on preventing nulls in the schema and having the default be `''` ?

mattlindsey:
Seems to work now that I also replaced GITHUB_ACCESS_TOKEN on production with a valid token.
So this PR should be ready to go now.

mattlindsey:
There is improved performance (less errors) using the oauth keys.
But now I'm thinking the problem is that that too many 'client' connections are created by this method, one for each call to Oktokit:

```
def client
    @client ||= Octokit::Client.new(:client_id => Settings.github.client_id, :client_secret => Settings.github.client_secret)
end
```

freeranger:
@tansaku do you have a timetable for when this will appear in prod?

tansaku:
@freeranger it's there I think

freeranger:
@tansaku Yes thanks it's working for me now.

PR #1658, opened 2017-05-12
Change get started now button link

### LINKED ISSUE #1652: "Get Started Now..." button should go elsewhere
Initial signup button should go to page with info about Premium membership, not just to page with only the cost (https://www.agileventures.org/premium)

mattlindsey:
fixes #1652


tansaku:
thanks @mattlindsey but I don't think this is doing what we want. See my inline comments

app/views/layouts/_escalating_call_to_action.html.erb
```diff
@@ -7,7 +7,7 @@
         <% message = 'Upgrade to Premium Mob for group sessions with a Mentor' %>
     <% end %>
 <% else %>
-    <% path = new_user_registration_path %>
+    <% path = '/premium' %>
```
tansaku: this is for when people are not logged in - we don't want to promote premium here - we want the `new_user_registration_path` - this shouldn't change

app/views/subscriptions/new.html.erb
```diff
@@ -1,8 +1,6 @@
 <article>
   <h1>Agile Ventures <%= @plan.name %> <%= type %></h1>
   <br/>
-  <a href='/<%= @plan.name.downcase.gsub(' ','_') %>'>Further details</a>
```
tansaku: this shouldn't be removed - we still need to link to further details here

tansaku:
What we need to change is this:

```
<% if current_user %>
    <% if current_user.membership_type == 'Basic' %>
        <% path = new_subscription_path %>
        <% message = 'Upgrade to Premium for additional support' %>
    <% elsif current_user.membership_type == 'Premium' %>
        <% path = new_subscription_path(plan: 'premiummob') %>
        <% message = 'Upgrade to Premium Mob for group sessions with a Mentor' %>
    <% end %>
<% else %>
```
to become this:

<% if current_user %>
    <% if current_user.membership_type == 'Basic' %>
        <% path = '/premium' %>
        <% message = 'Upgrade to Premium for additional support' %>
    <% elsif current_user.membership_type == 'Premium' %>
        <% path = '/premium_mob' %>
        <% message = 'Upgrade to Premium Mob for group sessions with a Mentor' %>
    <% end %>
<% else %>

tansaku:
Does that make sense?  Maybe we should pair on this if still unclear ...

mattlindsey:
@tansaku I think this PR is fixed now according to your comments and our hangout discussion.

tansaku:
excellent - thanks @mattlindsey ! :-)

Issue #1640, opened 2017-04-28
Paypal upgrade to Premium Mob not working


kirangp:
If a user is already enrolled to Premium subscription using Paypal, he/she is not able to upgrade the membership to Premium Mob using Paypal. If the 'Upgrade to Premium Mob' button is clicked, error 'No such customer: 5LERZVNJXBAR8' comes up.

mtc2013:
@alexander-kh has given a better error message for this condition.

this issue is closed by https://github.com/AgileVentures/WebsiteOne/pull/1659

PR #1659, opened 2017-05-12
Add handlers to fail gracefully for PayPal and other non-Stripe upgrades; fix issue #1670

alexander-kh:
#### Description

1. Added test that reproduces scenario described in the issue #1640
2. Add handlers to fail gracefully for PayPal and other non-Stripe upgrades

#### Motivation and Context

The goal is to make able to upgrade plan from `Premium` to `Premium Mob` by the PayPal.

Added corresponding acceptance test.
The test passes, that is, reproduces the message described in the issue.

#### Types of changes

#### Checklist:


tansaku:
an idea here @alexander-kh is that if this is complicated to fix, perhaps we start by releasing a change that just alerts the user to the problem rather than showing them the 500 error?

alexander-kh:
@tansaku, I think we can tackle this issue together.
Let me share my findings.

**Task**

The user has to be able to upgrade subscription plan from Premium to Premium Mob by PayPal from the profile page.
We have to adjust `update` method in `subscriptions_controller.rb`.
The Stripe track is configured according to [documentation](https://stripe.com/docs/subscriptions/upgrading-downgrading#switching).

**Brief introduction**

To implement PayPal services we've chosen PayPal Payments Standard solution, Subscribe payment button.

*Digression*
> We have to encrypt the payment button code.
> Button created in the PayPal account is protected automatically.
> As we created it manually, we have to hide the payment details so they cannot be seen by anyone viewing the website source code in a browser.
[Secure your HTML payment buttons](https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/encryptedwebpayments/)

*Implementation details*

Subscribe button submits the form (app/views/subscriptions/new.html.erb)
To upgrade the plan we have to retrieve transaction details in order to fetch subscription id.
Technically, PayPal provides `return` and `rm` variables that can be set up the way that the buyer's browser is redirected to the return URL by using the POST method, and all payment variables are included.
By handling HTTP response we can fetch `subscr_id`.
The question is in implementation.
PayPal also provides two notification methods: Instant Payment Notification (IPN) and Payment Data Transfer (PDT).
[IPN listener code sample](https://github.com/paypal/ipn-code-samples/blob/master/ruby/paypal_ipn_rails.rb)
Both methods have disadvantages.
- IPN messages can be delayed, but the service automatically resends messages for up to 4 days until the listener acknowledges them.
- PDT notifies immediately when a customer completes payment, but it sends confirmation only once.

*Another approach*

We already have `AddSubscriptionToUserForPlan` service (app/services/add_subscription_to_user_for_plan.rb).
Can we use it for the subscription plan upgrade?
The `update` method in `SubscriptionsController`. The code solves the task. It can be refactored and generalized for both payment methods.
```ruby
  def update
    if current_user.subscription.payment_source.is_a? PaymentSource::PayPal
      AddSubscriptionToUserForPlan.with(
        current_user, Time.now, Plan.find_by(third_party_identifier: "premiummob"),
        current_user.subscription.payment_source)
    else
      customer = Stripe::Customer.retrieve(current_user.stripe_customer_id)
      subscription = customer.subscriptions.retrieve(customer.subscriptions.first.id)
      subscription.plan = "premiummob"
      subscription.save
      current_user.subscription.plan = Plan.find_by third_party_identifier: 'premiummob'
      current_user.save
    end
    rescue Stripe::StripeError => e
      flash[:error] = e.message
      redirect_to user_path(current_user)
  end
```

tansaku:
thanks @alexander-kh - great documentation - I'm not sure I understand the difference between the two approaches, and it sounds like the digression is something we should open as another ticket maybe?

What concerns me is that another user might try to upgrade and get a 500 error before we can fix this - so perhaps we stamp out a quick fix to catch the 500 error and display a more helpful message to the user, and then we can consider alternatives at our leisure ...?

alexander-kh:
@tansaku, I've added redirection to the 'Premium Mob' sign up page, where the user will be able to upgrade plan either via PayPal or by credit card via Stripe.

app/controllers/subscriptions_controller.rb
```diff
@@ -32,15 +32,17 @@ def update
...
-    subscription.plan = "premiummob"
-    subscription.save
-    current_user.subscription.plan = Plan.find_by third_party_identifier: 'premiummob'
-    current_user.save
+    
+    plan = Plan.find_by(third_party_identifier: 'premiummob')
+    if user.subscription.payment_source.is_a? PaymentSource::Stripe
+      upgrade_plan_through_stripe(current_user, plan)
+    else
+      redirect_to new_subscription_path(plan: plan.third_party_identifier)
```
tansaku: I'm worried about redirect premiums to this page - there's a danger they sign up for Premium Mob and buy a second subscription - if we had something to tell them it was there responsibility to cancel their original Paypal account - and also CA grads would end up here ... but this is better than nothing I guess - maybe we need to make handling that part of another ticket and get this merged ... if it happens we can quickly get in there to request they shut down the original paypal subscription, or give CA's a discount etc ...

tansaku:
@alexander-kh thanks for these updates - btw, are you getting these build failures locally?

alexander-kh:
@tansaku, I haven't. Let me inspect them.

PR #1698, opened 2017-06-24
add cucumber test for join now patch 

### LINKED ISSUE #1695: ensure failing for the right reason acceptance test for join now issue
#1686 needs an acceptance test

lokeshh:
Fixes #1695


features/hangouts/edit_hangout_url.feature
```diff
@@ -30,4 +30,12 @@ Feature: Manual Edit of Hangout URL
...
     Then "Repeat Scrum" shows live for that hangout link for the event duration
     And "Repeat Scrum" is not live the following day
-    And there should be exactly 3 hangouts
\ No newline at end of file
+    And there should be exactly 3 hangouts
+
+  Scenario: Event doesn't go live before Hangout URL is updated
+    Given the date is "2014 Feb 5th 6:00am"
+    And I manually set a hangout link for event "Repeat Scrum"
+    Given the date is "2014 Feb 5th 7:05am"
```
lokeshh: @tansaku I had a hard time writing the cucumber test because it was also passing with the fix in #1694. I then realized that bug only happens unless [`@recent_hangout`](https://github.com/AgileVentures/WebsiteOne/blob/1eaccb5782e02325bd6797962bc94037111f9deb/app/controllers/events_controller.rb#L13) is not nil. In real application it doesn't stay nil but in the test it was staying nil. I couldn't understand what exactly is required to make it not nil. So I ended up adding lines line 36 and 37 above and line 115 in features/step_definitions/hangout_steps.rb

alexander-kh:
@lokeshh, good work!
I think we can make the language more clear.
* Scenario title
We can clarify the cause-and-effect relationships.
For example,
```
Scenario: The hangout link should be changed to the actual one at the time the event begins
```
* Flexible step definitions
For example,
```ruby
Then(/^there should be exactly (\d+) hangout(?:|s) associated with "([^"]*)"$/) do
```
to write the steps grammatically correct:
```ruby
    And there should be exactly 3 hangouts
    And there should be exactly 1 hangout associated with "LS"
```
I'll look at the actual changes tomorrow to complete the review.
A small note:
```ruby
time = Time.parse(@jump_date) + event.duration.minutes + 10.minute
```
This will work, but `10.minutes` will be read better.

lokeshh:
@alexander-kh Thanks for review!

I have following concerns-

1. `Scenario: The hangout link should be changed to the actual one at the time the event begins` This is not what's really happening. The hangout link has to edited manually and the scenario is to make sure that "Join Now" doesn't appear unless link has been updated.

2. Flexible step definitions. Yes that's good to have but I'm not sure if this is will be the correct PR to make such changes. I will change the `minute` to `minutes` at least because I wrote that.

alexander-kh:
@lokeshh, you are right. It works this way. The link has to be copied and pasted manually.
The technical part is correct. Good job!
But the git style can be improved from my perspective (you may leave it as is).
It's a good practice to group small changes into one descriptive commit, to grasp the essence of the changes by reviewing git log.
Do consider [this](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message) approach, for example. I hope it will be at least interesting reading.

features/hangouts/edit_hangout_url.feature
lokeshh: @tansaku Is these lines ok because I'm not sure on the logic how does this thing makes `@recent_hangout` not nil? Almost seems like magic. I think I'll be able to better explain this during a hangout when you will return.

alexander-kh: @lokeshh, when I remove lines 36 and 37 in `edit_hangout_url.feature` and line 115 in `hangout_steps.rb` the tests pass.
Note: You can test instance variables using `byebug` by adding step `Then I debug`

lokeshh: > From my perspective, we can leave it.

I'm confused what does "it" refer here? Do you mean the three lines which I've added?

lokeshh:
> Do consider this approach, for example. I hope, at least, it will be an interesting reading.

I agree I have to improve my commit style. I'll try to have better commit next time.

features/hangouts/edit_hangout_url.feature
alexander-kh: @lokkesh, yes, I mean your solution works and can be applied.
You're very good identified the problem.
[This](https://gist.github.com/alexander-kh/577166444f4e35d2647837aa01f6dca8) is another approach.

lokeshh:
Thanks to @alexander-kh suggestion [here](https://gist.github.com/alexander-kh/577166444f4e35d2647837aa01f6dca8) the cucumber test is simplified and much more readable with no magic code. @tansaku you can review and merge it now.

features/step_definitions/hangout_steps.rb
```diff
@@ -117,10 +118,26 @@
...
+  time = Time.parse(@jump_date) + event.duration.minutes + 10.minutes
+  Delorean.time_travel_to(time)
+  visit event_path(event)
+  expect(page).not_to have_link('Join now')
+end
+
+When(/^I start the "([^"]*)" event$/) do |event_name|
+  event = Event.find_by_name(event_name)
+  visit event_path(event)
+  find('#liveHOA-placeholder').click
```
tansaku: so this corresponds to clicking on the start hangout button? I don't think that ever caused the event to go live per se.

The event would go live when the hangout plugin in the hangout sent data back to the WSO server and that would make the event go live

tansaku:
as discussed in the scrum, I think we need to re-start this (or discard based on the fact that jitsi permalinks sort of make this issue redundant)

anyway, here are the notes from that chat:

1. tests to run fast
2. tests to be reliable
3. tests to catch bugs
4. tests to be quick to put together



1) Event would go live at a particular time, e.g. 7:00 UTC
2) Site Displays a link to "JOIN NOW" to the old hangout URL
3) Member would join the old URL
4) Scrum Master would submit a new Hangout URL
5) Scrum Master and others would join new Hangout

problem - people are in two different hangouts 

steps 1 and 2 should give us cucumber steps like so:

a) Given the date is "2014 Feb 5th 7:01am" (assuming event starts at 7:00am)
b) I should not see "Join Now" (check old Hangout URL not displayed)

Ironically jitsi permalinks resolve this issue for us

https://meet.jit.si/MartinFowlerScrum

----------------
Clicking on the hangout button in the tests has no effect o the WSO Rails Server
----------------

find('#liveHOA-placeholder').click  --> looking for hangout button

"JOIN NOW" link is what we need to actually check

tansaku:
video from that chat: https://youtu.be/qCruCItTc_0

features/step_definitions/hangout_steps.rb
alexander-kh: @tansaku, and this doesn't happen when clicking on "Start Hangout", right? This is complicated to figure out, as the hangout doesn't start in the development environment.
Known, that event can be started by filling in the `hangout_url` field.
The part of the logic in the `events/show.html.erb`, that increasing number of moving parts.
Since you are replacing the conferencing platform, apparently, this test should be reconsidered.
Essentially, the first approach of @lokesh solves the task. The intention was to make scenario clearer.

features/hangouts/edit_hangout_url.feature
lokeshh: @tansaku @mattlindsey Need your feedback. Please see my first comment above to see the problem.

mattlindsey: @lokeshh I think lines 36-37 are perfectly valid, especially since I think you should add the words "to a different URL" to the second to last step in the scenario, as I say in Slack.
I'm not sure what @tansaku thinks, but I think then you will have a perfectly good test for the problem that is happening.

mattlindsey: To clarify, add "to a different URL" to line 40 and add the step (with possible refactoring to make it more DRY), then test that the scenario fails on the original/unfixed code base, then test that it passes with the fix to the models.

lokeshh: @mattlindsey Yeah after adding the line "to a different URL" these lines 36-37 make sense. But `line 115 in features/step_definitions/hangout_steps.rb` still stands out odd.

mattlindsey: @lokeshh I guess it takes a page refresh to make 'JOIN NOW' appear on the page.  I can't think of a solution except possibly the one suggested here: https://robots.thoughtbot.com/automatically-wait-for-ajax-with-capybara

tansaku: @mattlindsey @lokeshh just revisit the page with `visit`

tansaku:
So I think a more comprehensible scenario might be:

```gherkin
  Scenario: Event doesn't go live before Hangout URL is updated
    Given that "Repeat Scrum" went live the previous day
    Then it should not go live the next day just because the event start time is passed
    When I manually set a hangout link for event "Repeat Scrum"
    Then "Repeat Scrum" shows live for that hangout link for the event duration
```

features/step_definitions/hangout_steps.rb
```diff
@@ -130,3 +141,20 @@
   expect(page).not_to have_content('This event is now live!')
 end
 
+Given(/^that "([^"]*)" went live the previous day$/) do |name|
+  steps %Q{
```
tansaku: we've tried to avoid nested steps due to problems with using the debugger and redirection

I guess we let that go here for the time being, since I think we're all tired of working on this one - would be nice to have a refactoring ticket perhaps ...

PR #1699, opened 2017-06-25
[NOT FOR MERGING] Show failing test for 1698

lokeshh:
Related to #1698 and #1694 

This PR is to show that new cucumber test added in #1698 fails for the right reason unless patched with #1694

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -222,28 +222,20 @@ class Event < ActiveRecord::Base
   def within_current_event_duration?
     after_current_start_time? and before_current_end_time?
   end
-  
-  def current_start_time
-    schedule.previous_occurrence(Time.now)
-  end
 
-  def current_end_time
-    schedule.previous_occurrence(Time.now) + duration*60
-  end
+  private
 
   def before_current_end_time?
-    Time.now < current_end_time
+    Time.now < (schedule.previous_occurrence(Time.now) + duration*60)
   rescue
     false
   end
 
   def after_current_start_time?
-    Time.now > current_start_time
+    Time.now > schedule.previous_occurrence(Time.now)
   rescue
     false
   end
-
-  private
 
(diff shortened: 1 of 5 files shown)


tansaku:
so the failure we have here is:

```
    When I start the "Repeat Scrum" event                                         # features/step_definitions/hangout_steps.rb:131
      Unable to find css "#liveHOA-placeholder" (Capybara::ElementNotFound)
      ./features/step_definitions/hangout_steps.rb:134:in `/^I start the "([^"]*)" event$/'
      features/hangouts/edit_hangout_url.feature:37:in `When I start the "Repeat Scrum" event'
```

so the question becomes is this the correct error.  It might be, but I'd love us to be really clear on this.

the absence of the #liveHOA-placeholder means that the start hangout button is not visible, which in the past was true more than ten minutes before the event and when the event has gone live ...

alexander-kh:
@tansaku, for reliability we can [add the time](https://gist.github.com/alexander-kh/577166444f4e35d2647837aa01f6dca8/revisions) in this particular step definition.
```ruby
When(/^I start the "([^"]*)" event$/) do |event_name|
   event = Event.find_by_name(event_name)
   Delorean.time_travel_to(Time.parse(@jump_date))
   visit event_path(event)
   find('#liveHOA-placeholder').click
 end
```

tansaku:
so this looks like the right error message:

```
    Then "Repeat Scrum" doesn't go live                                           # features/step_definitions/hangout_steps.rb:143
      expected not to find link "Join now", found 1 match: "JOIN NOW" (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/hangout_steps.rb:146:in `/^"([^"]*)" doesn't go live$/'
      features/hangouts/edit_hangout_url.feature:39:in `Then "Repeat Scrum" doesn't go live'
```

PR #1701, opened 2017-06-29
Update users show view spec

### LINKED ISSUE #1558: add cuke for #1551
this is almost so trivial I am not sure we need a test, but it was reported by a user in #1495 - it's a simple fix in the view - so I'm going with a separate ticket to merge in the fix - and leave it to anyone who's feeling motivated to reflect further on whether a test is necessary or what it should look like

sngeth:
This addresses the missing code coverage referenced here: https://github.com/AgileVentures/WebsiteOne/issues/1558

fixes #1558

I updated the view spec for the display user email link to ensure it is a proper mailto href. I believe this unit test is sufficient here as a high level feature seems overkill as it does not exercise much user interaction.


tansaku:
hi @sngeth - thanks so much for this - great work.

do say hi in our Slack chat if you're around :-)

Issue #1679, opened 2017-05-26
Clicking on videos in lists not working


mattlindsey:
For example, see http://localhost:3000/projects/websiteone and the first video appears, but when you click on any video in the list it doesn't appear.

Also see http://localhost:3000/users/sam-joseph and the first video does not appear, nor do the ones in the list which say:
video unavailable ('Start Broadcast' not pressed, or Hangout/YouTube fail)

tansaku:
once we fix the http/https issue we get another issue whith ic :

![](https://www.dropbox.com/s/zw89lgms8wj7fjh/Screenshot%202017-05-26%2016.54.00.png?dl=1)

lokeshh:
`video unavailable ('Start Broadcast' not pressed, or Hangout/YouTube fail)` is because `event_instance.yt_video_id` is nil and that's probably because we lost hangout API.

tansaku:
@lokeshh @mattlindsey so I wonder if this is fixed now ...

tansaku:
fixed now :-)

PR #1702, opened 2017-06-29
Fix click on videos not working

lokeshh:
Partial fix for #1679 

It works for videos which have link but doesn't work for videos which show the error "video unavailable ('Start Broadcast' not pressed, or Hangout/YouTube fail)"


app/assets/javascripts/users.js
```diff
@@ -4,7 +4,7 @@ WebsiteOne.define('Users', function () {
     function selectVideo(event) {
         event.preventDefault();
         player = $('#ytplayer');
-        player.attr('src', 'http://www.youtube.com/v/' + this.id + '?version=3&enablejsapi=1');
+        player.attr('src', 'https://www.youtube.com/embed/' + this.id + '?version=3&enablejsapi=1');
         $('#video_contents').text($(this).data('content'));
```
lokeshh: The problem was `www.youtube.com/v/` was returning 404 while `www.youtube.com/embed/` works.

tansaku:
great work @lokeshh !!!

tansaku:
wonder if we need a javascript test to wrap this?  or is this redundant since we can only really test manually ...? In the meantime I'll get this on production ...

PR #1708, opened 2017-07-06
manual update of youtube link

### LINKED ISSUE #1385: manual update of youtube link
with failure of google hangout api button, youtube link is no longer updated - one solution is to add a mechanism to update it manually

lokeshh:
Fixes #1385 #1679

### DIFF
--- a/app/controllers/event_instances_controller.rb
+++ b/app/controllers/event_instances_controller.rb
@@ -9,7 +9,6 @@ class EventInstancesController < ApplicationController
     hangout_url_changed = event_instance.hangout_url != event_instance_params[:hangout_url]
     yt_video_id_changed = event_instance.yt_video_id != event_instance_params[:yt_video_id]
     slack_notify = params[:notify] == 'true'
-
     if event_instance.try!(:update, event_instance_params)
       SlackService.post_hangout_notification(event_instance) if (slack_notify && event_instance.hangout_url?) || (event_instance.started? && hangout_url_changed)
       SlackService.post_yt_link(event_instance) if (slack_notify && event_instance.yt_video_id?) || yt_video_id_changed
@@ -71,7 +70,7 @@ class EventInstancesController < ApplicationController
         hangout_participants_snapshots_attributes: [{participants: params[:participants]}],
         participants: merge_participants(event_instance.participants, params[:participants]),
         hangout_url: params[:hangout_url],
-        yt_video_id: params[:yt_video_id],
+        yt_video_id: EventInstance.extract_yt_id(params[:yt_url]),
         hoa_status: params[:hoa_status],
         url_set_directly: params[:url_set_directly],
         updated_at: Time.now,
--- a/app/controllers/event_instances_controller.rb
+++ b/app/controllers/event_instances_controller.rb
@@ -70,7 +70,7 @@ class EventInstancesController < ApplicationController
         hangout_participants_snapshots_attributes: [{participants: params[:participants]}],
         participants: merge_participants(event_instance.participants, params[:participants]),
         hangout_url: params[:hangout_url],
-        yt_video_id: EventInstance.extract_yt_id(params[:yt_url]) || params[:yt_video_id],
+        yt_video_id: YouTubeRails.extract_video_id(params[:yt_url]) || params[:yt_video_id],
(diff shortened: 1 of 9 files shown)


features/hangouts/edit_youtube_url.feature
```diff
@@ -0,0 +1,31 @@
...
+      | Scrum        | Daily scrum meeting | Scrum    | 2014 Feb 4th 7am | 15       | never   | UTC       |                       |                                           |  LS     |
+      | Repeat Scrum | Daily scrum meeting | Scrum    | 2014 Feb 3rd 7am | 15       | weekly  | UTC       | 1                     | 31                                        |         |
+    And the following event instances (with default participants) exist:
+      | title        | hangout_url         | created_at       | updated_at          | uid | category | project    | user_id | yt_video_id | hoa_status | url_set_directly | event        |
+      | HangoutsFlow | http://hangout.test | 2012 Feb 4th 7am | 2012 Feb 4th 7:04am | 100 | Scrum    | Websiteone | 1       | QWERT55     | started    | true             | Repeat Scrum |
+      | HangoutsFlow | http://hangout.test | 2014 Feb 4th 7am | 2014 Feb 4th 7:03am | 100 | Scrum    | Websiteone | 1       | QWERT55     | started    | true             | Repeat Scrum |
+    And I am logged in
+
+  Scenario: Edit Youtube URL and ensure URL is saved
+    Given the date is "2014 Feb 4th 6:59am"
```
mattlindsey: Is there a reason to do one thing at 6:59 and the next at 7:01?  
If you're checking this, should we also say that the Event is scheduled to start at 7:00?  Or is that a different scenario  -- maybe the previous one you worked on?
@tansaku What do you think?

mattlindsey: Oh. I do see that you have a 7am event, so should the first step say "As a person involved in an event that starts at 7am"?

lokeshh: > Oh. I do see that you have a 7am event, so should the first step say "As a person involved in an event that starts at 7am"?

I've included that change. Thanks.

> Is there a reason to do it twice, at 6:59 and 7:01?
I'll make both the times 7:01 I'm doing it twice to test two different scenarios.

lokeshh:
@tansaku This is ready for review. All tests are passing. I haven't tested Slack notification functionality though.

app/controllers/event_instances_controller.rb
```diff
@@ -71,7 +70,7 @@ def transform_params(event_instance)
         hangout_participants_snapshots_attributes: [{participants: params[:participants]}],
         participants: merge_participants(event_instance.participants, params[:participants]),
         hangout_url: params[:hangout_url],
-        yt_video_id: params[:yt_video_id],
+        yt_video_id: EventInstance.extract_yt_id(params[:yt_url]),
```
tansaku: this seems like it would break existing functionality that relies on using the yt_video_id param ...

lokeshh: How? I don't think there's any existing functionality that updates `yt_video_id`.

lokeshh: @tansaku Seems like you were right... there were some tests requiring that param. I've modified the code now so it shouldn't be a problem now.

app/models/event_instance.rb
```diff
@@ -20,6 +20,18 @@ class EventInstance < ActiveRecord::Base
 
   validate :dont_update_after_finished, on: :update
 
+  def self.active_hangouts
+    select(&:live?)
+  end
+
+  def self.extract_yt_id(yt_video_url)
```
tansaku: so this new method would ideally have a rspec unit test to check that it was working ...

tansaku: I also wonder if we do this anywhere else - or if there is a library for it that would be even more generic?

I see three here: https://www.google.co.uk/search?q=extract+youtube+id+from+url

maybe https://github.com/luizpicolo/youtube_rails?

lokeshh: @tansaku Yes, I've added the specs for these. `youtube_rails` looks great and cover more variation of youtube video URLs. Should I use that? One argument against using it is that it will extra overhead.

PR #1723, opened 2017-07-11
adds recaptcha for email sign ups

### LINKED ISSUE #1067: Re-enable email sign-ups?
We had to disable this functionality recently due to spam-bots. Perhaps we could bring it back and make use of [Google reCaptcha](https://www.google.com/recaptcha/intro/index.html)

tansaku:
fixes #1067


tansaku:
So this passes locally - and I was able to sign up via a google auth, implying that that the captcha does not break non-email sign ups ...

tansaku:
@mattlindsey @mtc2013 @lokeshh so I'm actually surprised how this isn't breaking anything.

Seems like the point of recaptcha is to avoid happy path tests working https://github.com/ambethia/recaptcha/issues/232  I guess to unspike this we just need a sad path test that checks that robots are blocked from sign up ...

tansaku:
I guess the current `features/users/sign_up.feature` sign up features are all working because javascript is turned off ...

PR #1725, opened 2017-07-12
show error message for scrums with youtube id nil

### LINKED ISSUE #1719: avoid displaying videos in "scrums" when youtube id is null
https://www.agileventures.org/scrums shows lots of videos that won't play as it shows scrums when video ids are null (I think)

we have #1385 to try and allow videos to be updated going forward, but ideally scrums without a video id should not appear in this list

lokeshh:
Fixes #1719 

Removes the hyperlink from invalid youtube videos and display an error message


lokeshh:
Here's how it looks currently

image

features/scrums.feature
```diff
@@ -25,4 +25,9 @@ Feature: Scrums Index
     Then I should see a modal window with the first scrum
     When I close the modal
     And I click the second scrum in the timeline
-    Then I should see a modal window with the second scrum
\ No newline at end of file
+    Then I should see a modal window with the second scrum
+
+  Scenario: Videos with youtube id nill do not show
```
arun1595: Nit. How about _Videos with nil youtube id do not display youtube embed_ ?

lokeshh: That was the aim of this PR earlier but because project page and user profile page (https://www.agileventures.org/projects/websiteone#video) are showing en error so to be consistent we @tansaku @mattlindsey and me decided to display an error message.

lokeshh: Ah! Sure... Thanks

tansaku:
fantastic work here @lokeshh !!! :-)

PR #1733, opened 2017-07-17
Make sure duplicate event instances are not created

### LINKED ISSUE #1732: ensure that manual update of youtube link does not create new eventinstance
seems like updating youtube link manually causes a new event instance to be created?  would be great to avoid that ...

lokeshh:
Fixes #1732


app/views/events/show.html.erb
```diff
@@ -98,7 +98,7 @@
 
 <% if user_signed_in? %>
 
-    <% hangout_is_live = @hangout.try!(:live?) %>
+    <% hangout_is_live = @recent_hangout.try!(:live?) %>
```
lokeshh: The bug is caused by `@hangout` being always nil because its nowhere initialized. My guess is that it is a mistake by author which rather meant to write `@recent_hangout` and not `@hangout`.

tansaku: interesting

features/step_definitions/hangout_steps.rb
```diff
@@ -182,4 +185,7 @@
   page.should have_field('hangout_url', with: @hangout_url)
 end
 
-
+Then(/^a separate event instance is not created$/) do
+  expect(EventInstance.where("created_at >= ?", Time.now.beginning_of_day).size)
```
tansaku: I worry slightly that this test is dependent on the current time - if the test suite runs over the midnight boundary it could pass for the wrong reason.  Could we hard code the date based on the date set at the beginning of the cuke test?

lokeshh: I think the date coded at the beginning of the cuke test will persist to this step. Let me check using byebug.

lokeshh: I just checked, yes the date set in first step persist, so it doesn't matter when the test is ran. Time.now would return 2014-02-07 07:01:41 +0000

tansaku: yeah, I guess you're getting beginning of day - which is almost constant there, unless the test took a day to run, and that's a silly corner case, so I guess we're okay, but in general you should avoid using Time like that in a test

lokeshh: ok

PR #1747, opened 2017-07-20
added specs for associations, included modules and used shoulda instead of expect syntax

### LINKED ISSUE #1741: improving user specs
@praveen was suggesting

praveendhawan:
fixes #1741

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -9,13 +9,35 @@ describe User, type: :model do
   end
   subject { build_stubbed :user }
 
-  it { should have_one :subscription }
+  context "included modules" do
+    it { expect(User.ancestors.include?(Filterable)).to eq(true) }
+  end
 
-  it { is_expected.to have_many(:status) }
+  context "associations" do
+    it { should have_one(:subscription).autosave(true) }
 
-  it { is_expected.to accept_nested_attributes_for :status }
+    it { should have_one :karma }
 
-  it { is_expected.to respond_to :status_count }
+    it { should have_many(:authentications).dependent(:destroy) }
+    
+    it { should have_many(:projects) }
+    
+    it { should have_many(:documents) }
+    
+    it { should have_many(:articles) }
+
+    it { should have_many(:event_instances) }
+
+    it { should have_many(:commit_counts) }
+
+    it { should have_many(:status) }
+
+    it { should have_many(:status) }
+  end
+    
+  it { should accept_nested_attributes_for :status }
+
+  it { should respond_to :status_count }
 
   it 'should have valid factory' do
     expect(FactoryGirl.create(:user)).to be_valid


spec/models/user_spec.rb
```diff
@@ -9,13 +9,35 @@
...
 
-  it { is_expected.to have_many(:status) }
+    it { should have_many(:event_instances) }
 
-  it { is_expected.to accept_nested_attributes_for :status }
+    it { should have_many(:commit_counts) }
+
+    it { should have_many(:status) }
+
+    it { should have_many(:status) }
```
tansaku: is this a duplicate?

spec/models/user_spec.rb
```diff
@@ -9,13 +9,35 @@
   end
   subject { build_stubbed :user }
 
-  it { should have_one :subscription }
+  context "included modules" do
+    it { expect(User.ancestors.include?(Filterable)).to eq(true) }
```
tansaku: this reads a little strange - maybe we can't do anything about it, but could we have:

```rb
it 'includes Filterable module' do
  expect(User.ancestors).to include(Filterable)
end

**Comment by tansaku** [comment]
thanks for this @praveendhawan - good to have these in - hopefully it might help us think about the different roles and responsibilities of our User model better

**Comment by tansaku** [comment]
I thought Avdi had an article on how often "User" might sensibly replaced by "Account" but I can't seem to find it ...

File: spec/models/user_spec.rb [hunk 3]
```diff
@@ -9,13 +9,33 @@
   end
   subject { build_stubbed :user }
 
-  it { should have_one :subscription }
+  it "includes Filterable module" do
+    expect(User.ancestors).to include(Filterable)
+  end
+
+  context "associations" do
+    it { should have_one(:subscription).autosave(true) }
```
diraulo: I would use the `expect` syntax instead of `should`. You could rewrite the following as

```ruby
it { is_expected.to have_one(:subscription).autosave(true) }
```

spec/models/user_spec.rb
```diff
@@ -9,13 +9,33 @@
...
+
+    it { should have_one :karma }
```
diraulo: and this one would be

```ruby
it { is_expected.to have_one :karma }
```

tansaku:
great feedback @diraulo 

@praveendhawan does that make sense to you?

spec/models/user_spec.rb
praveendhawan: @diraulo @tansaku thanks for the feedbacks, will be updating it and will let you know.

praveendhawan:
@diraulo @tansaku I have updated the PR. Please have a look at it.

tansaku:
thanks @diraulo @praveendhawan :-)

PR #1752, opened 2017-07-27
add ClientMeeting event type and test

### LINKED ISSUE #1750: Client meeting event type
When creating an event the current options are: Scrum and PairProgramming. For the project also make sense to have ClientMeeting as a specific type.

mattlindsey:
fixes #1750


tansaku:
thanks for this @mattlindsey - sorry I should have reviewed earlier - looks like it's ready to go eh?

tansaku:
should we also be including tests that involve selecting this option with a feature specification that details why this new functionality is wanted, e.g. 

```
As a project manager
So that our members can distinguish client meetings from scrums and pair programming
I would like to create events that are categorised as client meetings
```

and we checked and found there was no other places in the code relying on classifying events in the original pair of categories?

tansaku:
@mattlindsey I've gone ahead and done the checks and adjusted the acceptance tests as I was suggesting - apologies if you were hoping to work on this, but I thought that given that it's been open in the PRs for a month I should just get it cleared up.

Very happy to go through it with you when you have more time :-)

PR #1753, opened 2017-07-27
[WIP] automate streaming of youtube video in jitsi

### LINKED ISSUE #1727: make it simpler to stream video from Jitsi
At the moment it's a little complicated to start a jitsy youtube stream

It would be great to have a button (in the jitsy) to start one, or even some set up that started them automatically - a button probably best in first instance

lokeshh:
Fixes #1727

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,7 @@ gem 'will_paginate-bootstrap'
 gem 'yt'
 gem 'yui-compressor'
 gem 'youtube_rails'
+gem 'google-api-client', '0.8'
 
 group :test do
   gem 'capybara'
--- a/app/assets/javascripts/events.js
+++ b/app/assets/javascripts/events.js
@@ -41,12 +41,12 @@ var jitsi_events = {
                 url: $(this).attr('href'),
                 timeout: 30000,
                 success: function(res) {
-                    alert(res.stream_name);
+                    alert('Start jitsi live stream with this key: ' +
+                        res.stream_name);
                 },
                 error: function(xhrObj, textStatus, exception) {
                     alert('Error!'); 
                 }
-                // 'success' and 'error' functions will be passed 3 args
             });            
         })
     }
--- a/app/assets/javascripts/events.js
+++ b/app/assets/javascripts/events.js
@@ -30,4 +30,14 @@ $(document).ready(function () {
     events.addToCalendar();
     editEventForm.handleUserTimeZone();
     showEvent.showUserTimeZone();
+    $('#btn-jitsi').click(function() {
+        var options = {
+            domain: "meet.jit.si",
+            room: "JitsiMeetAPIExample",
+            width: 700,
+            height: 700,
+            htmlElement: document.querySelector('#meet')
+        }
+        var api = new JitsiMeetExternalAPI(options.domain, options);        
+    })
 });
--- a/app/assets/javascripts/events.js
+++ b/app/assets/javascripts/events.js
@@ -33,6 +33,22 @@ var jitsi_events = {
             iframe.src = iframe.getAttribute('hidden_src');
             iframe_container.style.display = 'block';       
         })
+    },
+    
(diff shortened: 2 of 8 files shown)


app/views/events/show.html.erb
```diff
@@ -33,9 +35,10 @@
           </ul>
         </div>
 
-        <% if !@recent_hangout.try(:live?) and @event.less_than_ten_till_start?%>
-          <div class='btn-jitsi pull-right'>
-            <%= link_to @event.jitsi_room_link, target: '_blank' do %>
+        <% #if !@recent_hangout.try(:live?) and @event.less_than_ten_till_start?%>
+        <% if true %>
+          <div id='btn-jitsi' class='pull-right'>
```
lokeshh: Reminder to myself... remember to uncomment this.

Gemfile
```diff
@@ -63,6 +63,7 @@ gem 'will_paginate-bootstrap'
 gem 'yt'
 gem 'yui-compressor'
 gem 'youtube_rails'
+gem 'google-api-client', '0.8'
```
mattlindsey: Can you try using latest version?
See https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md

lokeshh: @mattlindsey That would require lot of time because I was able to pull up this one from existing examples given here https://github.com/youtube/api-samples/tree/master/ruby but there are no such examples for newest version.

mattlindsey: The upgrade doc explains it pretty clearly and I think @tansaku would agree it should be done so as not to use an outdated version. I can certainly help too.

mattlindsey: I do, however agree that it is a problem that google has not updated their samples. 
https://github.com/youtube/api-samples/issues/63#issuecomment-318918531

mattlindsey: And I am a little worried that this 'ruby api' for google is in 'maintenance mode' (according to the README.md: https://github.com/google/google-api-ruby-client#library-maintenance) and that we should be using the youtube api directly, but if it works I guess it is OK.

app/services/live_stream_service.rb
```diff
@@ -0,0 +1,123 @@
...
+  YOUTUBE_API_VERSION = 'v3'
+  
+  def self.get_authenticated_service
+    client = Google::APIClient.new(
+      :application_name => $PROGRAM_NAME,
+      :application_version => '1.0.0'
+    )
+    youtube = client.discovered_api(YOUTUBE_API_SERVICE_NAME, YOUTUBE_API_VERSION)
+  
+    file_storage = Google::APIClient::FileStorage.new("#{$PROGRAM_NAME}-oauth2.json")
```
mattlindsey: Can this be done using our standard method of having the api_key in environment variables?  Having this 'oauth2.json' file is a different method and would take administrative overhead to keep it working.

app/assets/javascripts/events.js
```diff
@@ -25,9 +25,38 @@ var events = {
     }
 };
 
+var jitsi_events = {
```
mattlindsey: This is cool.  It turns out you can also make ajax calls just using the `remote: true` flag on a button:  https://blog.codeship.com/unobtrusive-javascript-via-ajax-rails/

Gemfile
lokeshh: > but if it works I guess it is OK.

@mattlindsey So are you suggesting the old youtube api (0.8) is fine?

mattlindsey: @lokeshh I'll defer to @tansaku.  I would rather see it updated with latest ruby api, or converted to strictly using just youtube api.

tansaku:
@lokeshh should I close this for the time being? what remains to be done?

lokeshh:
@tansaku Yes you can close it for now. This PR will take a while to complete.

nisevi:
Closing this PR due to last comment.

Issue #1754, opened 2017-07-31
develop failing on test that previously past


tansaku:
So develop was green, but this test starting failing recently:

```
cucumber features/hangouts/edit_youtube_url.feature:33
```

Here's part of the stack trace:

```
    When I am on the "Show" page for project "LS"                                          # features/step_definitions/basic_steps.rb:255
      One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

      pause@https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:6480:34
      JV@https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:3410:51
      uV@https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:3383:182
      bra@https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:4146:126
      r1@https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:4125:103
      https://www.youtube.com/yts/jsbin/player-vflrwQIQw/en_GB/base.js:7223:167
      https://www.youtube.com/yts/jsbin/www-embed-player-vfllwwkQI/www-embed-player.js:373:293

```

It looks to me like the caching files for VCR and puffing-billy were not checked in with the original test commit, which makes it seem like something has changed in the way that youtube is responding to the request to embed a video in a project page?

I notice I can remove the error if I remove the ytplayer iframe from the project page - I guess this might be to do with requesting something from a video that doesn't exist.  I say that because when I add that video id to an event associated with a project on the develop server, I get the following error in the js console:

![](https://www.dropbox.com/s/pkorlde2c8ejhdk/Screenshot%202017-07-31%2014.52.05.png?dl=1)

that appears to be a request for the default youtube video image or https://i1.ytimg.com/vi/12341234111/hqdefault.jpg

and I'm not seeing that issue for general project videos on production ...

if I paste that URL into a browser I get: 
 ![](https://www.dropbox.com/s/q6or14smu0ivtu5/Screenshot%202017-07-31%2014.57.42.png?dl=1)

the ytplayer is updated by code in app/assets/javascripts/users.js:

```js
    function selectVideo(event) {
        event.preventDefault();
        player = $('#ytplayer');
[...]

PR #1755, opened 2017-07-31
disable this test until we can figure out the issue better

tansaku:
related to #1754


tansaku:
@lokeshh @mattlindsey we should merge this to ensure develop is green again, but please do look through first to see if it's clear what the issue is here ...

mattlindsey:
@tansaku I agree that we should put this in for now.  I sort-of understand the issue.

PR #1767, opened 2017-08-08
[WIP] replace mercury editor with medium_editor for rails

### LINKED ISSUE #1756: Replace mercury editor in rails 4
In order to upgrade to rails 5, it may be better to replace mercury with something first.

I imagine that the mercury rails engine is not compatible with Rails 5. Perhaps we might try switching to a simple markdown editor like https://github.com/rderoldan1/md_simple_editor (we saw convoluted test fails all over the place)

mattlindsey:
fixes #1756


tansaku:
Well done getting this started @mattlindsey

app/assets/javascripts/static_pages.js
```diff
@@ -0,0 +1,16 @@
...
+                placeholder: {
+                    /* This example includes the default options for placeholder,
+                        if nothing is passed this is what it used */
+                    text: 'Type your text',
+                    hideOnClick: true
+                }
+            }
+        };
+    }
+});
```
mattlindsey: @tansaku Need help. How do I get this javascript to be called on init of every 'static page'?

tansaku: You could call it later in that same file, although I'm not sure about that define function.

If I recall almost all the controllers seem to use slightly different approaches to JavaScript in this project.

There's also something weird about JavaScript being loaded on only the relevant pages or on all of them. I've needed to add extra calls in the views on occasion.

Does the medium rails gem not specify?

mattlindsey: The medium rails gem does not specify.  I'm thinking of trying this approach: http://brandonhilkert.com/blog/page-specific-javascript-in-rails/

tansaku: Sure, although I'd recommend doing the simplest thing possible first just to check it works - I was just looking at the docs - why not just put the JavaScript tag in the view to start with?

nisevi:
Hi there guys @mattlindsey @tansaku, I been taken a look to the changes and also to what is implemented now and I think that I would need the guidance of any of you in order to tell me how is `mercury` intended to work, and where is being impelmented.

After that I think that I would have a better idea of how to do the next step by implementing `medium`.

My personal opinion in the matter is that we should not be using the editor for the actions that we are performing, but is only based in what I have been able to understand. I think that it would be better to implement javascript with ajax calls when we want edit specific parts of a `project`.

One of the issues we currently have with the implementation of mercury is this one:

image

The steps for reproducing this bug are:

- hit the edit button on a project;
- hit the change section button and select an option;
- press the edit button again;
- you will be able to open as many mercury editors as section on the project;

In my opinion, by removing the mercury editor and by presenting a normal form we will remove the bug and we will be able to focus on things like styles,  upgrade to rails 5, and ajax calls for updating attributes individually, in order to make the site look modern.

I would love to hear your opinions :+1:

nisevi:
Hey @sigu what do you think about my last comment? any thoughts?

tansaku:
@nisevi the point of this PR is to remove mercury and replace it with medium - if mercury doesn't work we don't care - we want to rip it out and replace with the medium editor

the key thing is this spike is getting the medium editor working - that's what Matt and I had started on previously

nisevi:
@tansaku maybe I'm not being clear with my idea, I don't think we need an editor at all.

Being said that and if we still need to implement an editor, can we pair on this one? I think that I need to understand an little more the scope of this PR.

tansaku:
@nisevi one possibility would be to remove editing entirely and only allow updates to static pages via pull request, but that then excludes some people from editing and means the cycle for changing pages becomes longer ...

nisevi:
@tansaku closing this PR since we are gonna work on the GitHub integration for static pages.

Issue #1682, opened 2017-05-31
allow manual addition of videos


tansaku:
users often wonder how they can add videos to a project (had that from at least a couple of different sources).  Being able to manually set the video of an event to get round the loss of the Hangout API would be one solution here.

roschaefer:
same same. When I click on Videos I expect a big button with a plus saying "Add video". At the very least, an explanation or quick tutorial how to do so. Up to this point I don't understand and I cannot see an option how to add videos via hangouts (if at all this is the way how to populate the video list)

c1505:
I'll work on this.  From my understanding, it seems that this request is to modify the link to the video and potentially add new links and descriptions to videos rather than a way to upload and host the video ourselves?   

Also, do you know offhand if these missing videos are failures of the API  and/or just what shows up when a planned hangout does not take place?

![image](https://user-images.githubusercontent.com/6599077/28830688-814e411c-76a5-11e7-9284-9067c544a1cf.png)

tansaku:
great to have your input here @c1505 

> From my understanding, it seems that this request is to modify the link to the video and potentially add new links and descriptions to videos rather than a way to upload and host the video ourselves?

I don't think we support video "descriptions" at present.  The request is to give some sort of admin access to videos so that when a video is missing (due to whatever reason), someone who has the link to that video can come in and set the video link.

> Also, do you know offhand if these missing videos are failures of the API and/or just what shows up when a planned hangout does not take place?

If no one starts the hangout, then nothing shows up.  It's API failure - basically because the google hangouts API just doesn't work any more

@lokeshh added new functionality that allows the youtube URL to be set when the event starts, so it can be set manually via the event, while it is running so to speak, but there is currently no functionality to set it later, which is what this ticket is about.

It's possible (in principle) to update the video associated with an event_instance via the rails console, but there are a load of validations in place that make it really hard.

I almost think that the way to approach this issue is to start by creating a show page for individual event_instances so that they can be inspected individually through the web interface and then move on to allow for editing them in various ways ...

PR #1770, opened 2017-08-15
SPIKE edit youtube link for EventInstance

c1505:
I want to see if this is the desired path before I go too far along.  Appropriate tests, links to the edit page, and authorization will be added before submitting this to merge.    

#### Description
Spike for issue #1682 .  Adds the ability to edit a youtube link for an EventInstance.  

#### Questions
* Are videos hosted in places other than youtube?  
  * Currently this path would only be for youtube videos.  The youtube video ID is what is persisted so I am sticking to that in order to not break things elseware.  
* Is there currently any role based authorization?
  * One of the things mentioned in the issue was to have the link editable by an admin.  At first glance,
  I didn't see any roles for users.  I think it would make sense to have that.  I would also have the link 
 modifiable by the creator of the associated event.

### DIFF
--- a/app/controllers/event_instances_controller.rb
+++ b/app/controllers/event_instances_controller.rb
@@ -36,12 +36,13 @@ class EventInstancesController < ApplicationController
 
   def update_link
     @event_instance = EventInstance.find(params[:id])
-    if @event_instance.update_attributes(yt_video_id: YouTubeRails.extract_video_id(event_instance_params[:yt_video_id]) )
+    youtube_id = YouTubeRails.extract_video_id(event_instance_params[:yt_video_id])
+    if youtube_id && @event_instance.update_attributes(yt_video_id: youtube_id )
       flash[:notice] = "Link Updated"
       redirect_to edit_event_instance_path(@event_instance)
     else
       flash[:alert] = "Error.  Please Try again"
-      render edit_event_instance_path(@event_instance)
+      redirect_to edit_event_instance_path(@event_instance)
     end
   end
 
--- a/app/controllers/event_instances_controller.rb
+++ b/app/controllers/event_instances_controller.rb
@@ -1,6 +1,7 @@
 class EventInstancesController < ApplicationController
   skip_before_filter :verify_authenticity_token
-  before_filter :cors_preflight_check, except: [:index]
+  before_filter :cors_preflight_check, except: [:index, :edit, :update_link]
+  # before_action :authenticate_user!, only: [:edit, :update_link]
 
   def update
     event_instance = EventInstance.find_or_create_by(uid: params[:id])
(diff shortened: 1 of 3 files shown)


tansaku:
thanks @c1505 - great start.

In answer to your questions, at the moment videos are only hosted on youtube.  Supporting other video hosters would be a separate feature I think.

At the moment we have no role based authorization.  The entire AV site is based around the idea that every individual user has the same permissions for updating everything, a bit like wikipedia.

I've taken the approach over the years that we should make everything accessible to everyone and add roles etc. as problems emerge.  Ironically most people can't believe they do have access, don't make changes, and when they notice they can, they report it as a bug.  Looking back at it I'm not sure how much we've gained from the full open access approach.

I'm an advocate of simplicity first, or putting off technical decisions to the last minute (like Uncle Bob), so it still might make sense to add the feature without role based authorization first, or we could impose some restrictions.   I've been thinking that it might be sensible to move towards a stack overflow model where abilities within the site are a function of karma score, but again that should be another separate feature.

In the first instance we could deploy the feature without any role-based authorization, but not link to it, so only those who new the link could do the edit.  That might seem obtuse, but this is a heavy load bearing portion of the system, which is frustratingly not very RESTful (ideally all updates would take place through a single, coherent `update` method, and creation would take place via a `create method` but there are lots of dependencies) ... it would be good to deploy covertly and have in use by those who knew the link for a while (to fix up a few video issues) and see how it works with everything else before we follow up with a plan for authorization and linking into the main site?

Does that make sense? Apologies for length of post ... really great to have you contributing again :-)

c1505:
Makes sense!  Thanks for the quick response.

Do we want to have a separate index page with links to the edit pages or just the edit page?  What would be the way to find the ID of the EventInstance if there is no link to it?

Is there a preference to open a new PR with a new branch when I start writing the tests or should I update this one?

tansaku:
we could have a separate index page for the event_instances - I think that will be quite useful at some point, but note that we already have https://www.agileventures.org/hangouts and https://www.agileventures.org/scrums (note you can see in this one the places where we want videos to be able to be manually edited) and I was going to say we can grab the event_instance ids from there, but now that I look closely I see we don't show the ids ...

so maybe a separate (more traditional) list is in order ...  that might be easiest.  It will be great to see a generalized list view for admin purposes - which we can maybe ultimately replace https://www.agileventures.org/hangouts with once we've working out the pathway for all this ...

You are free to open a new PR - that allows you to easily see the tests fail by writing them first.  You could also stay working in this branch, but then you'll need to disable the functionality you already added in order to see your tests fail which takes more coordination and discipline.  Up to you ... :-)

tansaku:
@c1505 as it happens I've just been working on some performance and CSS issues for the /hangouts page.  

We've got the existing `index` command there:

```rb
  def index
    relation = (params[:live] == 'true') ? EventInstance.live : EventInstance.latest
    relation = relation.includes(:project, :event, :user)
    @event_instances = relation.paginate(:page => params[:page], per_page: 5)
  end

```

it would not actually be too difficult to add a link to the eventinstance show view from with the `views/event_instances/_index_header.html.erb` partial, and I think it would be pretty obscure ...

c1505:
Not quite sure what you mean here.  Assuming you meant a different partial?  I submitted a PR #1793 that just includes the ID as a data attribute so that it is exposed, but hidden if someone wants to inspect on it.  If you prefer something different, let me know.  Let's continue the discussion there :)

PR #1783, opened 2017-08-19
Update ubuntu_14.04_manual_install_notes.md

fybwid:
Update ruby version at [Generate rvm config files within project](https://github.com/AgileVentures/WebsiteOne/blob/develop/docs/ubuntu/ubuntu_14.04_manual_install_notes.md#generate-rvm-config-files-within-project)

#### Description

Update ruby version as noted in Gemfile

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/ubuntu/ubuntu_14.04_manual_install_notes.md
+++ b/docs/ubuntu/ubuntu_14.04_manual_install_notes.md
@@ -65,9 +65,9 @@ git clone git@github.com:<your github handle>/WebsiteOne.git
 ## Generate rvm config files within project
 ```
 cd WebsiteOne
-echo "ruby-2.2.2" > .ruby-version
+echo "ruby-2.3.1" > .ruby-version
 rvm gemset create websiteone
-rvm use ruby-2.2.2@agileventures
+rvm use ruby-2.3.1@agileventures
 echo "websiteone" > .ruby-gemset
 ```
 


FedericoEsparza:
So there's a bit of contradicting code that was here previously. Plus it's just a lot of commands, when it can all be done with one command, and the commands, currently, will cause error messages.
 
Instead of all of the following: 
`echo "ruby-2.3.1" > .ruby-version`
`rvm gemset create websiteone`
`rvm use ruby-2.3.1@agileventures`
`echo "websiteone" > .ruby-gemset`
 
@fybwid Yosef, could you please update the file, instead, with:
`rvm use ruby-2.3.1@WebsiteOne --ruby-version --create`
 
This updates the ruby version to 2.3.1, saves it to the`.ruby-version` file, creates the `WebsiteOne` gemset (notice that "One" is capitalized, to match the project name exactly), switches to 2.3.1 as the current ruby version, and **does not** try to use an uncreated `agileventures` gemset.
 
So that takes care of the error message related to the uncreated `agileventures` gemset. 

However, there will still be an error message, because in the **Install ruby** section just above this section the user is still being asked to install 2.2.2: `rvm install ruby-2.2.2`
 
So, Yosef, could you please also modify that section to: `rvm install ruby-2.3.1`?
 
**Thanks for your good work on this!**

PR #1746, opened 2017-07-20
Youtube URL is posted on Slack but not Hangout URL when Youtube URL is editted

lokeshh:
Fixes the bug that **both Youtube and Hangout URL are posted when Youtube URL is edited**.

Discussed in https://www.youtube.com/watch?v=IKtO9cn4Rbc

PR #1793, opened 2017-08-22
Edit past Youtube URL for EventInstance (fixes  #1682)

### LINKED ISSUE #1682: allow manual addition of videos

c1505:
Edit past Youtube URL for EventInstance (fixes  #1682)
See spike #1770 for additional discussion.  

#### Description
Adds the ability to edit a youtube link for an event_instance.  As discussed, there is currently no link to this edit page, but a hidden data attribute was added to the past hangouts page.


tansaku:
great work @c1505 - this is looking really good.

I'd love to completely sort out the REST interface for event_instances, but that's another ticket I think :-)

tansaku:
ah, but one issue - I think editing the event_instance takes the event live, which is something we need to avoid ... See the example here - the top two events have gone live (join button green)

screenshot 2017-08-24 18 22 15

c1505:
Thanks!  I agree it is a bit awkward that it is named as an additional action, but it didn't seem to belong with the existing update action with how much else is going on there unrelated to a vanilla update action.

tansaku:
@c1505 yeah, you did the right thing under the circumstances - it's a bigger refactor to sort out that other mess

c1505:
Ah I forgot the code to determine if it is live checks on when the event was updated.  I'll look at that

tansaku:
yeah, check that here: https://github.com/AgileVentures/WebsiteOne/blob/develop/app/models/event_instance.rb#L43

tansaku:
the pain here might be that we have to change the way an event_instance checks that it is live ...

tansaku:
or we need the ability to edit all the other fields - as I think I mentioned it might make sense to have a simple show page that lists all the fields for an event_instance and then allow us to edit at will ... that said an edit page with pre-filled boxes is good too - no immediate need for another route ...

c1505:
Pretty sure this last WIP commit will cause some tests to fail.

tansaku:
@c1505 I think it might be a simpler path to allow us to edit the `hoa_status` rather than changing the way the live status is represented ...

c1505:
Yea you are right.  I just added the hoa_status to the edit page and defaulted to "finished" .  

It seems more like a band-aid though.  Changing behavior based on the last time an EventInstance was updated seems like bad practice.  If we want ```.live?`` to be based on an update time maybe a separate well named object without additional fields that would potentially be edited manually for some reason in the future.  Maybe another possibility could be to have a reported_live field in an EventInstance with a DateTime and compare that to the current DateTime to determine if it is live.  

I am good with this change as it currently is though if you are.

tansaku:
Thanks, and I agree this is a band-aid.  However the code around this point is in flux.  There are question marks about how it's going to evolve.  I agree that ultimately we need to move away from checking live status based on updated_at, but the higher priority is to release this new piece of functionality and rule out any unintended side effects in the system.  Perhaps more importantly, making that change as part of this PR could bog us down and prevent us quickly evaluating this one small chunk of functionality.

Long and painful experience with legacy systems, and this system in particular, makes me ultra-cautious.

I think it's also the case that we don't necessarily want to default the status to "finished",  as if we do so and someone has an event live and happens to update the youtube video while the event is taking place it will force it into finished mode - could we perhaps have the hoa_status in the form default to whatever the current value is for the event_instance?

That then allows the admin to both understand the current status and adapt to it as necessary ...

We could create a refactoring ticket for changing the `live` check away from using `updated_at`, but I think the critical thing might be to focus on delivering the smoothest experience for folks trying to use hangouts and jitsys.  We have a bug at the moment that @lokeshh is working on a fix for:

https://github.com/AgileVentures/WebsiteOne/pull/1746

that we were discussing in the WSO meeting yesterday: https://youtu.be/73fYJPTYI4c

Notice that that PR is also looking at code in the same area that you are working on here.  There are always multiple refactorings we could do.  The key is to work out which are the set of refactorings to do that allow us to deliver the best experience to the user with the minimum of effort :-)

c1505:
I'll change it so that hoa_status doesn't default to "finished".  

My thinking was that this was going to be used to edit finished events and the person doing the editing wouldn't assume they have to change the hangout status in order to prevent an event from going live that previously was not.

c1505:
There was code that did not allow an EventInstance to be updated after the hoa_status was marked "finished".  I removed that code and the corresponding spec in the last 2 commits.  I don't see any code that relied on this behavior and having it prevents us from updating the youtube url if the EventInstance is marked as "finished".

tansaku:
thanks @c1505 - hopefully that won't blow up in our face :-)

PR #1798, opened 2017-08-25
fix previously turned off js tests

lokeshh:
Related to #1754 

I'm trying to turn on js on the tests which had to be turned off to made the develop build pass.

### DIFF
--- a/features/hangouts/edit_youtube_url.feature
+++ b/features/hangouts/edit_youtube_url.feature
@@ -39,7 +39,7 @@ Feature: Manual Edit of Youtube URL
     #
     # see https://github.com/AgileVentures/WebsiteOne/issues/1754
     #
-  @javascript_ignore_js_errors
+  @javascript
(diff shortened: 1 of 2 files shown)


lokeshh:
2 out of 3 tests which were failing earlier are now passing on my local machine.

lokeshh:
@tansaku As you can see 2 of the 3 tests are passing now which were failing earlier. But I have no clue why that's the case.

lokeshh:
@tansaku Also the 3rd test is passing now which was failing on my local computer like 3 days ago. So, we have got all the tests passing now which we had disabled earlier.

tansaku:
@lokeshh yeah, that's weird - something to investigate, however we are critically blocked by this issue at the moment: https://github.com/AgileVentures/WebsiteOne/pull/1746 can you prioritize resolving this above all else?  

many thanks

tansaku:
these are all passing for me locally now ... weird - safe to pull in I guess ...

PR #1818, opened 2017-09-14
Support multiple repositories for projects

### LINKED ISSUE #761: Multiple source repository
As a user
When my project has more then 1 repository
I would like the ability to define them in the project page

Copied from https://www.pivotaltracker.com/projects/982890/stories/108643006

tansaku:
fixes #761


app/assets/javascripts/projects.js
```diff
@@ -19,3 +19,16 @@ WebsiteOne.define('Projects', function() {
     }
   }
 });
+
+$(document).on('ready', function(){
```
tansaku: could be namespaced and jasmine tested ...?

tansaku: will do that as part of UX review

PR #1833, opened 2017-09-26
Empty github repo makes the github_commits_job run in infinite loop.

### LINKED ISSUE #1789: handle empty github project more gracefully
new project has empty github - caused rake task to hang:

> Waiting for Github to calculate project statistics for Lanternn/Unitcom

https://github.com/Lanternn/Unitcom

praveendhawan:
fixes #1789


tansaku:
@praveendhawan let me know if you need some help - maybe we could pair on this? https://www.agileventures.org/premium-f2f-offer

praveendhawan:
@nisevi Let me know if you are up for pairing session for this ticket. I am kinda stuck on this one.

nisevi:
hey @praveendhawan my internet connection is back again to normal :+1:  I would love to pair with you to understand a little more about this issue and how could I help you :+1:

nisevi:
Hi there guys @praveendhawan @tansaku , I was reading the octokit [documentation](https://octokit.github.io/octokit.rb/Octokit/Client/Stats.html#contributors_stats-instance_method) related to the method we are using for getting the contributor stats and I saw that the method accepts not only the the name of the repo but also a hash parameter that can be setup for using these options:

```
Parameters:

repo (Integer, String, Hash, Repository) — A GitHub repository
retry_timeout (Hash) — a customizable set of options
retry_wait (Hash) — a customizable set of options
```
This give us the freedom to remove the `loop` that we were using, and at the same time as we don't have the `loop` anymore, the configuration for the retries is no longer needed it.

If you wanna read more about it, you can take a look [here](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/stats.rb).

app/jobs/github_commits_job.rb
```diff
@@ -55,12 +55,9 @@ def update_user_commit_counts_for(project)
     end
 
   def get_contributor_stats(repo)
-    loop do
-      contributors = client.contributor_stats(repo)
-      return contributors unless contributors.nil?
-      Rails.logger.warn "Waiting for Github to calculate project statistics for #{repo}"
-      sleep 3
-    end
+    contributors = client.contributor_stats(repo, { retry_timeout: 6, retry_wait: 3 })
```
nisevi: @praveendhawan @tansaku as it can be seen [here](https://github.com/octokit/octokit.rb/edit/master/lib/octokit/client/stats.rb):

```
module Octokit
  class Client

    # Methods for the Repository Statistics API
    #
    # @see https://developer.github.com/v3/repos/statistics/
    module Stats

      # Get contributors list with additions, deletions, and commit counts
      #
      # @param repo [Integer, String, Hash, Repository] A GitHub repository
      # @option retry_timeout [Number] How long Octokit should keep trying to get stats (in seconds)
      # @option retry_wait [Number] How long Octokit should wait between retries.
      # @return [Array<Sawyer::Resource>] Array of contributor stats
      # @see https://developer.github.com/v3/repos/statistics/#get-contributors-list-with-additions-deletions-and-commit-counts
      # @example Get contributor stats for octokit
      #   @client.contributors_stats('octokit/octokit.rb')
      def contributors_stats(repo, options = {})
        get_stats(repo, "contributors", options)
      end
      alias :contributor_stats :contributors_stats
```

On the implementation of the method `contributor_stats` we can send those two parameters for making the retries and the time that it should wait making retries. I just did the setup for the request to `timeout` in 6 seconds, this means that the it will retry for a maximum period of 6 seconds. I also did the setup for how longer octokit should wait between retries to 3 seconds. This setup was totally arbitrary and based on what I considered it could be a good configuration, but I'm maybe wrong and I would really appreciate your opinion here :+1:

nisevi: here you have the definition for the method `get_stats`:

```ruby

      private

      # @private Get stats for a repository
      #
      # @param repo [Integer, String, Hash, Repository] A GitHub repository
      # @param metric [String] The metrics you are looking for
      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
      # @see https://developer.github.com/v3/repos/statistics/
      def get_stats(repo, metric, options = {})
        if retry_timeout = options.delete(:retry_timeout)
          retry_wait = options.delete(:retry_wait) || 0.5
          timeout = Time.now + retry_timeout
        end
        loop do
          data = get("#{Repository.path repo}/stats/#{metric}", options)
          return data if last_response.status == 200
          return nil unless retry_timeout
          return nil if Time.now >= timeout
          sleep retry_wait if retry_wait
        end
      end
```

nisevi:
@tansaku yes this will prevent the loop since the way the loop works within the gem is difference from what it was implemented, also if we want to test it we are gonna test the code inside the gem and not the behavior that we have in our code. IMO there is no chance to generate an infinite loop now.

nisevi:
@tansaku have you seen that we don't have a loop anymore in our code? and the loop that I'm sharing in the comments is the one that belongs to the gem? I'm just making this clarification because due to the comments in the review it may be confusing to follow along with the code :+1:

app/jobs/github_commits_job.rb
praveendhawan: nice findings @nisevi . Looks good to me as well @tansaku. But I would say we should have that test case for empty repo as well.

tansaku:
@nisevi yes, great fix - we're just waiting for @praveendhawan to add the tests to ensure that the fix can be tested automatically ...

praveendhawan:
@tansaku @nisevi Only thing remaining is updating VCR file, which I will update after getting updated `env` file.

tansaku:
@praveendhawan just sent that to you

PR #1873, opened 2017-10-08
[WIP] first pass at allowing customer to add a new credit card

### LINKED ISSUE #1852: allow sponsored users to add their credit card details if they want to take over payment
Just adding a card causes a problem as we assume an existing stripe id:

```
Stripe error while updating card info: Could not determine which URL to request: Stripe::Customer instance has invalid ID: nil for #<User:0x00....4818>
```

tansaku:
fixes #1852 

needs to have:

* [x] sad path test
* [ ] ~refactor stripe interactions to services?~


tansaku:
this seems to have fixed it: https://github.com/rebelidealist/stripe-ruby-mock#custom-errors

PR #1972, opened 2017-11-18
1960 Add sponsor relationship

### LINKED ISSUE #1960: Add sponsor-sponsee relationship
Add change to model to accommodate accurate representation of when a user sponsors another.

marianmosley:
- Add sponsor_id column to subscriptions table
- Set sponsor to current user in the subscriptions controller

fixes #1960

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -15,7 +15,7 @@ class SubscriptionsController < ApplicationController
 
     create_stripe_customer unless paypal?
 
-    add_appropriate_subscription(@user)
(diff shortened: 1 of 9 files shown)


mtc2013:
@tansaku if sponsoring user is not signed-in, which they are not currently required to be, then we have no way to properly record who they are in an automated fashion.  We could assume in this case that the user is being sponsored?  Or should we start requiring them to be logged-in first?

tansaku:
@mtc2013 I think it's time to require being logged in - please feel free to add that

tansaku:
code climate says:

mtc2013:
@tansaku after our discussion, is this ready to go?

PR #1999, opened 2017-12-05
support posting links to multiple channels

### LINKED ISSUE #1943: update as many channel project/slack room associations as possible
e.g. Y, AV Community Talk, Marketing etc.

tansaku:
fixes #1943

So here I've added the ability to ping multiple slack channels - and added a test to ensure that works for pair programming.

I've updated the channels so that the marketing meeting will ping the marketing channel and av_community channel and that the autograders meeting will also ping the betasaasers channel ...

I've also merged in the minor refactoring of using mock_model, but I'm worried that this is all getting a bit messy, and we need to maybe re-design pushing from TDD.  

I guess the key thing I want to move towards is supporting private events that would ping premium_extra - that's a bigger undertaking ...

also, the specs are rather difficult to follow, and maybe driving from Cucumber would make this all more comprehensible ...

### DIFF
--- a/app/services/slack_service.rb
+++ b/app/services/slack_service.rb
@@ -11,35 +11,37 @@ module SlackService
     return unless Features.slack.notifications.enabled
     return if hangout.hangout_url.blank?
 
-    channel = channel_for_project(hangout.project)
+    channels = channels_for_project(hangout.project)
 
     message = "#{hangout.title}: <#{hangout.hangout_url}|click to join>"
     here_message = "@here #{message}"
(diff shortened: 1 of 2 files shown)


tansaku:
ideally all the channel information would all be stored in the database and be configurable through the interface to attach particular projects to particular channels ..., but perhaps private events is the higher priority, and that could work with simple hard coding perhaps ...

what I'd love to see is have it so that premiums could see the linked videos for private events, but non-premiums could not, or could see only one, and then after that would be prompted to upgrade ...

PR #2012, opened 2017-12-12
adding a logged in requirement for premium subscriptions

### LINKED ISSUE #1976: require login to sponsor user or purchase plan for self
Right now people don't have to be logged in to hand over payment.

tansaku:
fixes #1976

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,5 +1,6 @@
 class SubscriptionsController < ApplicationController
 
+  before_action :authenticate_user!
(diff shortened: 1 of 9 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -51,6 +53,18 @@ def update
 
   private
 
+  def require_login?
+    true unless action_name == "create" && Rails.env.test?
```
tansaku: we could maybe make this even more specific here by checking that the request is `paypal?`

PR #2013, opened 2017-12-13
Project page displays links to all associated github repos.

### LINKED ISSUE #1839: display multiple repos for project when available
currently we only display the primary one

bcb37:
#### Description
Created and tested a change to the projects page template that renders the "connections" sidebar widget.

#### Motivation and Context
We want to display all of the information associated with a project on it's details page.

I created a new cucumber scenario in the features set for the page that shows the project details, along with an additional table to represent the source repositories in the database. I also created corresponding steps in the associated steps file. These tests passed, as did the full test suite for the application.

#### Types of changes

#### Checklist:

fixes #1839


bcb37:
I also noticed that I erroneously designated this a "new feature". It's probably more of a "breaking change"?

app/views/projects/_connections.html.erb
```diff
@@ -1,8 +1,10 @@
 <aside class="sidebar-widget projects-sidebar-group">
   <h1 class="widget-title">Connected on</h1>
-  <% if @project.github_url.present? %>
-    <p><i class="fa fa-github-alt fa-lg"></i>
-      <span class="small"><%= link_to "#{@project.github_url.split('/').last} ", @project.github_url %>on GitHub</span></p>
+  <% if (@project.source_repositories.size > 0) && ! @project.source_repositories[0].url.nil? %>
```
tansaku: I wonder if we could remove this line, and replace it with more defensive coding in the loop, or possibly with a helper, e.g. `repo.url.try(:split, '/').try(:last)`

alternatively a validation to ensure the URL is never nil?

features/step_definitions/projects_steps.rb
```diff
@@ -23,6 +23,15 @@
   end
 end
 
+Given(/^the following source repositories exist:$/) do |table|
+  table.hashes.each do |hash|
+    project = Project.find_by(title: hash[:project])
+    project.source_repositories.delete_all
+    source_repository = SourceRepository.new(url: hash[:url], project_id: project.id)
```
tansaku: or `project.source_repository.build(url: hash[:url])`

tansaku:
@bcb37 great work - I made a few comments on possible changes - happy to make them myself if you're keen to move on to other things :-)

app/views/projects/_connections.html.erb
bcb37: I checked for the nil url before entering the loop because otherwise it was possible to enter it without there being any repos in the db for the project (and so it wouldn't output "not linked to GitHub"). This was the only way I could get all the tests to pass. The "Github url (primary)" field on the edit page only seems to show up if there are `source_repositories` associated with the project. But when you create a new project on the site and don't fill in that field, it doesn't build a blank source repository automatically, like it does in the tests steps. In fact, the edit and create functions seem a little awkward as a result of this: If there are no repos, there's still a js button labeled "Add more repos", and clicking it is the only way to get the "Github url (primary)" field to appear. 

I added validation for URL on the `source_repository` model, and changed the tests to use @javascript and click the "Add more repos" button. Everything passed except the `migrate_github_urls` test (the one example with no github url was expecting "" and got nil instead). At this point it's probably becoming another issue. Should this one just be merged in as is?

tansaku: interesting that you say:

> But when you create a new project on the site and don't fill in that field, it doesn't build a blank source repository automatically

I thought I had set that up so that it did, but maybe I'm mistaken ... :-/

tansaku: when I go to `projects/new` in the main site I get an automatic repo box ...



ah, but so if we don't fill it in we end up with an associated source repo with a nil in the URL ...? hmm does start to sound like a new issue ...

tansaku: however I just tried the following code:

```erb
  <% if @project.source_repositories.size > 0 %>
      <% @project.source_repositories.each do |repo| %>
        <p><i class="fa fa-github-alt fa-lg"></i>
        <span class="small"><%= link_to "#{repo.url.try(:split,'/').try(:last)} ", repo.url %>on GitHub</span></p>
      <% end %>
  <% else %>
```

and all the project tests are green.  I also note we have a load of code in the Project model that would make more sense in the SourceRepository model now, i.e. 

```rb
  def github_repo
    if github_url.blank?
      ''
    else
      /github.com\/(.+)/.match(github_url)[1]
    end
  end

  def github_repo_name
    github_url ? /github.com\/(\w+)\/\w+/.match(github_url)[1] : ''
  end

  def github_repo_user_name
    github_url ? /github.com\/\w+\/([\w\-]+)/.match(github_url)[1] : ''
  end
```

although arguably your method is cleaner in the case that github url has more slashes ... could just pull that into SourceRepository ...?  I love `try` but it leaves me conflicted.  I feel like code should be robust to data inaccuracies ... (Avdi Grimm's Confident Ruby) ... maybe a method on SourceRepository like so:

```rb
class SourceRepository < ActiveRecord::Base
  belongs_to :project
  def name
    return '' if url.nil? 
    return url unless url.contains '/'
    url.split,'/'.last
  end
end
```

then we could have:

```erb
  <% if @project.source_repositories.size > 0 %>
      <% @project.source_repositories.each do |repo| %>
        <p><i class="fa fa-github-alt fa-lg"></i>
        <span class="small"><%= link_to "#{repo.name} ", repo.url %>on GitHub</span></p>
      <% end %>
  <% else %>
```

Anyhow, we don't want to get off track fixing up other shortcomings in the system, but we're supposed to minimize the amount of code in the views that we could pull into models (that said I have some glaring offences like that in other parts of the codebase ...).

If you're keen to spend any more time on this then be my guest, but if you're ready to move on to another ticket I'll make some final tidies and pull in, but very happy to chat in more detail if you can join any scrum :-)

bcb37: I'm ok with pulling it in at this point. There are certainly big-picture concerns here that I hadn't thought of. I'll bring a couple brief questions to an upcoming scrum.

Issue #1989, opened 2017-11-24
Google sign up not working?


tansaku:
One user reported an error trying that method

tansaku:
user in question confirms they saw this error:

![](https://dl.dropbox.com/s/vj1c7er18cpi4ho/Screenshot%202017-11-27%2010.43.04.png?dl=0)

tansaku:
I can confirm that I cannot currently login with google auth ...

tansaku:
I just tried updating the production secret - didn't help - something else going on here I think ...

tansaku:
maybe something needs to change in: https://github.com/AgileVentures/WebsiteOne/blob/develop/app/controllers/authentications_controller.rb

nisevi:
@tansaku closing this since its already fixed :+1:

PR #2053, opened 2018-01-03
1989_omniauth_google_sigin_work_in_progress

theBoyMo:
#### Description

An update in the omniauth gem resulted in Google signin failing for users. The failure was not picked up by any of the current tests.

related to #1989

#### Motivation and Context

Since sign in is such an important service, it would be useful to try and implement a test that could possibly pick this up in the future.

I have not been able to implement a test that would pick up the specific problem. I have added two scenarios that test the sad path, a user signing in with invalid credentials via google or github. This is a work in progress. Although the google scenario passes, the Github one does not. The error message generated states that Capybara could not find "with Github" button.

#### Types of changes

#### Checklist:


features/users/omniauth.feature
```diff
@@ -8,14 +8,29 @@ Feature: Connect to social sites
...
 
+  @omniauth-with-invalid-credentials
+  Scenario: Try to log in with invalid credentials
+    When I click "with Google"
+    Then I should see "invalid_credentials"
+    And I should be signed out
+
+  @omniauth-with-invalid-credentials
+  Scenario: Try to log in with invalid credentials
+    When I click "with Github"
```
tansaku: @theBoyMo I think the capitalization might need to change here, i.e. "GitHub"

tansaku:
great work @theBoyMo - will be lovely to add these two sad path tests ...

features/users/omniauth.feature
theBoyMo: Your right, everything passes once I added the correction.

theBoyMo: I had one question, the error message displayed is "invalid_credentials", since that is the symbol used, and not "Invalid credentials". Is that a problem? I was assuming not since we're just testing that we have a sad path in place.

PR #2097, opened 2018-01-13
[WIP] Cucumber scenarios: one pass (Event is live), one fail (countdown)

### LINKED ISSUE #1820: Several hours after event goes live, it still shows as live even though long over
User complained about staleness of event when they did not reload page.

We could use the length of time of the event to approximate when it is likely to have ended.

bcb37:
#### Description

fixes #1820 
#### Types of changes

#### Checklist:


bcb37:
The problem here I think has to do with time travel (using Delorean) only working on the rails server, not the phantomjs one. Seems like, to make this acceptance test work, we'd need to also get javascript's Date() function to return the traveled-to date. The only solutions I've found that do that involve putting a test in the layout templates, which alters the actual application code.

tansaku:
@bcb37 thanks for this - I seem to remember encoutering a similar issue in the main events page where we have a js thing display the event time in the time zone of the browser.  I think the work around we ultimately used there was to check that the setup HTML in the view was the correct for the js to interpret, but that we couldn't achieve a complete end to end test.  It might be the same here ...

tansaku:
I guess the other option here is to insert javascript to mock the date function that phantomjs calls, but so that in the test instead of the application code, however we don't want to burn too much time on it if it's gets complicated

tansaku:
I guess you might have already looked at http://metaskills.net/2012/08/07/synchronized-times-in-ruby-and-javascript-acceptance-tests-using-capybara-timecop-and-sinon-js/

bcb37:
I've tried looking into ways of doing this without modifying the app, but none have worked so far. The current commit puts the code in the head template. I waded a bit into the interface between configuring test/dev/production environments and the asset pipeline (especially with respect to how and when js files get cached). The solution from the link you posted seems to suggest that it's possible to cordon off the test stuff in an 'integration' area of the assets directory (and maybe configure the app so those things are ignored by sprockets for production builds). But I couldn't make that work. Also, I know it's probably a bad idea to include the lolex build in the repo, but I wasn't sure what the best strategy would be to avoid that.

bcb37:
This seems to work and pass all tests. The countdown now continues to update every second, even after the event has gone live, until the duration has run out (well, really until 59 seconds before that, because it's testing whether the minute is 0), after which it reports that it has ended and stops caring. There is also the matter of where to get lolex, since it shouldn't be packaged with the app the way it is.

bcb37:
Thanks Sam! That scenario wording makes sense. And I suppose I should have changed the title of this PR before I added the final, non-WIP commit. Are there any further steps for me on this one?

PR #2114, opened 2018-01-20
1865 - Upgrade pivotal tracker API

nisevi:
https://waffle.io/AgileVentures/WebsiteOne/cards/59d72bf1b5729a0024e8aaa5


nisevi:
@tansaku the code coverage has decreased but it doesn't seems to be related to the changes I've done. Or at least I'm not able to find that out on coveralls site, if you could point me in the rights direction I'll increase the code coverage :+1:

PR #2116, opened 2018-01-21
Allow event creator to select audience for an event (front-end only)

### LINKED ISSUE #1596: Private events for premium members
```
As a premium member
So that I can easily work out the time of the next premium event in my timezone
I would like premium events

As an admin
So that we can encourage premium signups
I'd like premium events to be only accessible to premium members (but potentially visible to non-premium members --> call to action to become premium member in order to join)
```

we could do with private events that were only visible to premium, mob, or f2f members as we are currently organizing those on an ad hoc basis

marianmosley:
fixes #1596


nisevi:
Hi there @marianmosley, could you update the description of the PR with the link to the issue to know a little more about it?

bcb37:
The state of this WIP seems to be stable. The 'for' field is present after migration and appears on the CRUD templates. All tests pass. The app is not using the flag for anything yet. Should that make it mergeable as a first complete slice of the issue?

PR #2121, opened 2018-01-23
1849 - Upgrade factory girls

nisevi:
https://waffle.io/AgileVentures/WebsiteOne/cards/59d33836ebe9e900c6d59398

There are tests failing due to this [after_filter](https://github.com/AgileVentures/WebsiteOne/blob/develop/app/controllers/application_controller.rb#L11) on ApplicationController, the specific message says:

```
      Failure/Error: current_user.try :touch
      
      RuntimeError:
        stubbed models are not allowed to access the database - User#touch()
      # ./app/controllers/application_controller.rb:70:in `user_activity'
```


nisevi:
@tansaku this PR will upgrade to `factory_bot_rails` however I'm trying to see why we have this error regarding the `after_filter` I mentioned in the description.

spec/controllers/projects_controller_spec.rb
```diff
@@ -14,9 +14,10 @@
   #TODO split specs into 'logged in' vs 'not logged in'
   before :each do
     # stubbing out devise methods to simulate authenticated user
-    @user = build_stubbed(User, id: 1, slug: 'some-id')
+    @user = build_stubbed(:user, id: 1, slug: 'some-id')
     allow(request.env['warden']).to receive(:authenticate!).and_return(@user)
     allow(controller).to receive(:current_user).and_return(@user)
+    allow(@user).to receive(:touch).and_return(true)
```
nisevi: @tansaku here I intercepted the touch method that is being called in the after_filter on ApplicationController and I made it to return true. This is what I thought it could be a solution but not sure if it is the right approach, I would love to know what you think :+1:

spec/controllers/static_pages_controller_spec.rb
```diff
@@ -15,6 +15,7 @@
   before(:each) do
     allow(request.env['warden']).to receive(:authenticate!).and_return(user)
     allow(controller).to receive(:current_user) { user }
+    allow(user).to receive(:touch).and_return(true)
```
nisevi: @tansaku I did the same here :+1:

nisevi:
@tansaku I think that I solved it, but I would like to have your opinion on some review comments I have created :+1:

spec/controllers/projects_controller_spec.rb
tansaku: yeah, this is fine, although looking at the code that gets run in application controller:

```rb
  def user_activity
    current_user.try :touch
  end
```
you don't have to return a value I think, or you could even just adjust the user like so:
```rb
@user = build_stubbed(:user, id: 1, slug: 'some-id', touch: true)
```
not much difference, but this latter would make the test slightly shorter I think ...

nisevi: @tansaku I followed your advice and I tried:

`@user = build_stubbed(:user, id: 1, slug: 'some-id', touch: true)`

However I found this error:

```
Failures:

  1) ProjectsController#index #show renders the show template
     Failure/Error: @user = build_stubbed(:user, id: 1, slug: 'some-id', touch: true)
     
     NoMethodError:
       undefined method `touch=' for #<User:0x0055fec24a10e8>
       Did you mean?  touch
     # ./spec/controllers/projects_controller_spec.rb:17:in `block (2 levels) in <top (required)>'
```

So I decided to just not make it to return something, what do you think?

PR #2181, opened 2018-02-25
Adjust cucumber scenario to explicitly tweak the updated_at variable

### LINKED ISSUE #1939: Adjust Cucumber Scenario to explicitly tweak the `updated_at` variable
#1809 was an annoying bug that I think we've finally fixed, and we now have a cuke scenario that covers its two forms.  However to catch the issues with `updated_at` I changed the `updated_at` value in the background data and so in principle that could have knock on effects on the other scenarios in the feature.

Also, it would probably be better to make explicit mention in the bug fix scenario that we're trying to avoid more recent updates to an event cause it to hang around when it's a new day and we want to be starting fresh ...

This is the bug fix scenario:

https://github.com/AgileVentures/WebsiteOne/blob/develop/features/hangouts/edit_hangout_url.feature#L41

and this is where updated_at was changed:

https://github.com/AgileVentures/WebsiteOne/blob/develop/features/hangouts/edit_hangout_url.feature#L18

The main thing for this would be to identify if we had a good existing step to use to modify the updated_at on an event instance or if we need to write a new one ...

maryam21:
Adjusted cucumber scenario to modify the updated_at variable.
#### Description

I added a step into 'Event doesn't ping old youtube URL' scenario so that it can modify the value of the updated_at variable.
#### Motivation and Context

This change is required so that we can test against a different updated_at value without modifying it in the background data.

fixes #1939
#### Types of changes

#### Checklist:


tansaku:
many thanks for this @maryam21 - much appreciated! :-)

maryam21:
Thanks :)

PR #2183, opened 2018-02-26
'puffing-billy' gem upgrade.

nisevi:
https://github.com/AgileVentures/WebsiteOne/pull/2171


nisevi:
The problem we are having is con `coveralls` since it is [downgrading our code coverage](https://coveralls.io/builds/15702592) and I trying to figure our why.

With Dependabot it was happening the same, so I created this PR thinking that it was a problem with the dependabot upgrade, but it seems that it was not.

PR #2203, opened 2018-03-08
add rundfunk slack channel

### LINKED ISSUE #2190: Saving an Event's YouTube link pings Hangout Slack message
**Still posts properly** to the **project's video section**, however:

tansaku:
might fix #2190


tansaku:
really should be adding a test for what happens when we don't have slack channels specified, at least I thought we had those, but maybe they are not quite testing what I think any more ... but pushed for time, want to get this out to see if it helps in production or not ...

Issue #2209, opened 2018-03-09
is_privileged? on User model should handle when Settings.privileged_users is nil.


nisevi:
On model `WebsiteOne/app/models/user.rb` this method should handle when `Settings.privileged_users` is `nil`

```
  def is_privileged?
    Settings.privileged_users.split(',').include?(email)
  end
```

PR #2210, opened 2018-03-09
Handle nil on Settings.privileged_users

nisevi:
Handle `nil` when `Settings.privileged_users` doesn't exists.

Fixes https://github.com/AgileVentures/WebsiteOne/issues/2209

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -155,6 +155,7 @@ class User < ApplicationRecord
   end
 
   def is_privileged?
+    return false if Settings.privileged_users.blank?
     Settings.privileged_users.split(',').include?(email)
   end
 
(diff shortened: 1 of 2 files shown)


nisevi:
@tansaku  with this we are handling when `Settings.privileged_users` is not setup.

nisevi:
@tansaku the ticket was already created, it usually appears in this way for you to check it next time :+1: :

image

I will add the `fixes #ticket_id` as you said and the unit test for when `Settings.privileged_users` is not setup.

nisevi:
@tansaku our codecoverage ir showing us that we already cover that piece of code:

image

However this is the case because we are returning `false` if this `Settings.privileged_users` is not setup. So I will add a new individual test for testing when this setup returns `nil`.

nisevi:
@tansaku I have done all the mentioned changes :+1:

PR #2254, opened 2018-03-21
Upgrade ruby to latest version.

nisevi:
#### Description
Upgrade to ruby version 2.5.0 for having the [performance improvements](https://www.ruby-lang.org/en/downloads/) that come with the upgrade.


nisevi:
@tansaku @sigu as we talked in the #websiteone-devops channel I'm merging this in.

Issue #2004, opened 2017-12-06
[Airbrake] [Production] PQconsumeInput() SSL error: decryption failed or bad record mac : SELECT "event_instances".* FROM "event_instances"


diraulo:
**Airbrake error:** [#8543](https://herokuapp21498474herokucom.airbrake.io/projects/117119/groups/2099408349284528543)
**Airbrake project:** websiteone-production

**Error type:** `ActiveRecord::StatementInvalid PG::ConnectionBad`
**Error message:** `` PQconsumeInput() SSL error: decryption failed or bad record mac : SELECT "event_instances".* FROM "event_instances" ``
**Where:** `rake#karma_calculator`
**Occurred at:** Dec 06, 2017 15:09:37 UTC
**First seen at:** Dec 06, 2017 15:09:37 UTC
**Occurrences:** 0 (0 since last deploy on `<no information>`)
**Severity:** `error`

**URL:** `<no information>`
**File:** `[GEM_ROOT]/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb`

**Backtrace:**
```
[GEM_ROOT]/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in async_exec
[GEM_ROOT]/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in block in exec_no_cache
[GEM_ROOT]/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_adapter.rb:484:in block in log
[GEM_ROOT]/gems/activesupport-4.2.10/lib/active_support/notifications/instrumenter.rb:20:in instrument
...
Caused by PG::ConnectionBad: PQconsumeInput() SSL error: decryption failed or bad record mac
```

PR #2261, opened 2018-03-23
Query database once for all users

### LINKED ISSUE #2221: rake#karma_calculator
The Rake Karma Calculator task is a background task that reviews all the user data to calculate the karma score that is then used to rank them on the users page.

The rake task is located here `lib/tasks/scheduler.rake`:

```ruby
task :karma_calculator => :environment do
  User.all.each do |usr|
    KarmaCalculator.new(usr).perform
    usr.karma.save!
  end
end
```

This is the code of the class itself `app/services/karma_calculator.rb` and probably where the bug exists:

```ruby
class KarmaCalculator

  attr_reader :user

  def initialize(user)
    @user = user
  end

  def perform
    if user.karma
      user.karma.total = 0
    else
      user.karma = Karma.find_or_create_by(user_id: user.id, total: 0)
    end

    return if user.created_at.blank?
    user.karma.total = sum_elements
    # better to have time in pairing sessions, code contributed (related to quality), issues, ...
  end

  private

  def sum_elements
    membership_length + profile_completeness + activity + number_hangouts_started_with_more_than_one_participant + number_github_contributions + hangouts_attended_with_more_than_one_participant
  end

  def membership_length # 6
    user.membership_length
  end

  def profile_completeness # 10
    awarded = user.profile_completeness
    awarded += user.authentications.count * 100
    return awarded
  end

  def number_github_contributions
    user.commit_count_total
  end

  def number_hangouts_started_with_more_than_one_participant
    user.number_hangouts_started_with_more_than_one_participant
  end

  def hangouts_attended_with_more_than_one_participant
    id = gplus_id
    return 0 unless id
    hangouts_attended = calculate_hangouts_attended_with_more_than_one_participant(id)
    user.hangouts_attended_with_more_than_one_participant = hangouts_attended
  end

  def gplus_id
    gplus_auth = user.authentications.select { |a| a.provider == 'gplus' }
    gplus_auth.try(:first).try(:uid)
  end

  def calculate_hangouts_attended_with_more_than_one_participant(id)
    EventInstance.all.select do |i|
      more_than_one?(i) && i.participants.to_unsafe_h.values.any? do |p|
        p['person']['id'] == id
      end
    end.count
  end

  def more_than_one?(event_instance)
    !event_instance.participants.nil? && event_instance.participants.to_unsafe_h.values.count > 1
  end

  def activity # 6
    user.activity
  end

end
```

It's been suggested that this isn't working correctly and may be related to the following airbrake issues: https://github.com/AgileVentures/WebsiteOne/issues/2206
https://github.com/AgileVentures/WebsiteOne/issues/2241
https://github.com/AgileVentures/WebsiteOne/issues/2245

NickSchimek:
#### Description
The database was being queried to grab all EventInstances for each user. The EventInstance query is the same for each user. So that is about 8000 repetitive queries to the database. Moved this query outside of the loop.

#### Motivation and Context
Reoccurring Airbrake issues related to the KarmaCalculator show that an exception is occurring and the process is killed. Issue #2004 in particular due to a bad PG connection and time out.

According to the Heroku website "Scheduled jobs are meant to execute short running tasks or enqueue longer running tasks into a background job queue. Anything that takes longer than a couple of minutes to complete should use a worker dyno to run.” Confirmed this is not being ran on a worker dyno.

Fixes #2221 

I tested the speed of this rake task on my system and it revealed that it would take anywhere between 1 and 6 seconds to process each User before the change. After the change it is taking less than a second per User.

Before changes: 
- Time elapsed for 100 users is: 236.5s (4 mins). 
- Estimate to run for 8000 users would be over 5 hours.

After changes:    
- Time elapsed for 100 users is: 49.1s (<1 min).
- Real time run for 8000 users was just over an hour.

This is a 5x increase in speed and reduces the load on the database from O(n) to O(1).

Further improvement can be made to increase the speed of the rake task which will reduce the run time to just a couple minutes.

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

#### Checklist:

### DIFF
--- a/app/services/karma_calculator.rb
+++ b/app/services/karma_calculator.rb
@@ -1,9 +1,10 @@
 class KarmaCalculator
 
-  attr_reader :user
+  attr_reader :user, :events
 
-  def initialize(user)
+  def initialize(user, events)
     @user = user
+    @events = events
   end
 
   def perform
@@ -55,7 +56,7 @@ class KarmaCalculator
   end
 
   def calculate_hangouts_attended_with_more_than_one_participant(id)
-    EventInstance.all.select do |i|
+    events.select do |i|
       more_than_one?(i) && i.participants.to_unsafe_h.values.any? do |p|
         p['person']['id'] == id
       end
--- a/app/services/karma_calculator.rb
+++ b/app/services/karma_calculator.rb
@@ -57,14 +57,9 @@ class KarmaCalculator
 
   def calculate_hangouts_attended_with_more_than_one_participant(id)
     events.select do |i|
-      more_than_one?(i) && i.participants.to_unsafe_h.values.any? do |p|
-        p['person']['id'] == id
-      end
+      participant = i.participants.to_unsafe_h.values
+      participant.count > 1 && participant.any? { |p| p['person']['id'] == id }
     end.count
-  end
-
-  def more_than_one?(event_instance)
-    !event_instance.participants.nil? && event_instance.participants.to_unsafe_h.values.count > 1
   end
 
   def activity # 6
--- a/lib/tasks/scheduler.rake
+++ b/lib/tasks/scheduler.rake
@@ -9,8 +9,9 @@ task :fetch_github_last_updates => :environment do
 end
 
 task :karma_calculator => :environment do
-  User.all.each do |usr|
-    KarmaCalculator.new(usr).perform
-    usr.karma.save!
-  end
+  events = EventInstance.all.where.not(participants: nil)
+    User.all.each do |usr|
+      KarmaCalculator.new(usr, events).perform
+      usr.karma.save!
+    end
 end
--- a/lib/tasks/scheduler.rake
+++ b/lib/tasks/scheduler.rake
@@ -10,8 +10,14 @@ end
 
 task :karma_calculator => :environment do
   events = EventInstance.all.where.not(participants: nil)
-    User.all.each do |usr|
+  User.all.each do |usr|
+    begin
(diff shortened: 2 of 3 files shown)


tansaku:
thanks @NickSchimek - this looks like a huge improvement

could we just throw that error handling line in here and fix the identing?

```rb
task :karma_calculator => :environment do
  events = EventInstance.all.where.not(participants: nil)
  User.all.each do |usr|
    try
      KarmaCalculator.new(usr, events).perform
      usr.karma.save!
    rescue e
      Rails.logger(e)
    end
  end
end
```

or maybe the error handling should be inside the KarmaCalculator?

Well anyway, no need to wait for that, want to get this deployed ASAP, but if you have time to throw in over the weekend before we look at it on Monday - either way :-)

Really appreciate your great progress here!

NickSchimek:
@tansaku Good catch on the indentation. I wasn't sure if you still wanted to throw in that error handling. I will do that now and I think we should wrap it just how you suggested.

NickSchimek:
Error handling has been added which should allow for better debugging if this error continues. Also I took some better measurements of the performance boost that will be gained from these changes and edited the post above with the results.
Again Heroku suggests that these rake tasks shouldn't take more than a couple minutes unless it's on a worker dyno. However this rake task can be improved more so that it only takes a couple minutes to run each night which will be the next task to do.

tansaku:
Now, the acid test will be getting this onto production and seeing if it fixes our issue

PR #2279, opened 2018-04-06
Decrease running time of karma_calculator rake task

### LINKED ISSUE #2221: rake#karma_calculator

NickSchimek:
#### Description
Update karma calculator to use User.event_participation_count. This will save hours of processing time since the data currently does not change.

Added rake task modify_event_participation which needs to be ran ONLY ONCE to update all users event_participation_count.

fixes #2221 
#### Motivation and Context
This is needed to prevent dyno's from timing out.

Previous rake task was taking 5hrs to run locally. The rake task now runs in 68 seconds.

#### Types of changes

#### Checklist:


app/services/karma_calculator.rb
```diff
@@ -43,27 +43,11 @@ def number_hangouts_started_with_more_than_one_participant
     user.number_hangouts_started_with_more_than_one_participant
   end
 
-  def hangouts_attended_with_more_than_one_participant
```
tansaku: should we be moving this code into a rake task so that when/if participation data gets fed in again we can run it as a separate one off task to update the csv file?

tansaku:
great start here @NickSchimek - see my inline comment

app/services/karma_calculator.rb
tansaku: you make a strong argument :-) as @nisevi knows I'm a big fan of keeping things around :-) but I guess there's always git history, and you're right, it's a very inefficient implementation and if we do get something new set up we should plan and implement that separately ... okay let's merge and get you working on the jitsi stuff - that's a huge thing for us that could really streamline the community ...

PR #2294, opened 2018-04-18
Notify project creator when new people join project

### LINKED ISSUE #1813: Notify project creator when new people join project
**As a** project organizer/creator, 
 
**I want to** be notified when new people join my project within the AV site, 
 
**so that** I can reach out and personally welcome them.

maryam21:
When a new member joins a project, an email is sent to notify project creator.
#### Description

1. Added cucumber test for the feature.
2.  Added model method to send email to project creator then called the method via projects controller follow action.
3. Added rspec unit tests for controller and model method. 
#### Motivation and Context

Fixes #1813 
#### Types of changes

#### Checklist:


tansaku:
thanks for starting on this @maryam21 - great work - were you also planning a cucumber acceptance test?

maryam21:
Thanks @tansaku :) I am not sure I got your question, but I just put the steps when the user interacts with the feature.

maryam21:
I think it's a random fail because just before the last of merging this branch with develop the build passed. I think it does exercise the same code but it's more specific it makes sure that the follow action actually calls alert_project_creator_about_new_member method, so if for example the if the acceptance test fails and this one passes we would know that the call is actually done but it's just that something went wrong somewhere else. But that's just my opinion I just though that it will make identifying any related problem more easily.

tansaku:
I think you're correct @maryam21 that it could help us pinpoint errors faster - also that needs to be offset against the cost of maintaining more code ... but let's merge this in - no need to delay taking things in or out I think ...

maryam21:
I get your point, I should have taken that into consideration as well.

Issue #2216, opened 2018-03-12
GDPR compliance


tansaku:
provide option to opt-in to newsletter - and if they don't then don't add them to elasticemail for newsletter

praveendhawan:
@tansaku where this option should be shown?

tansaku:
hi @praveendhawan - apologies for slow response - have been on vacation - this should be an option shown on sign up - there is currently no options at all - I think we need to add to:

https://github.com/AgileVentures/WebsiteOne/blob/develop/app/views/devise/registrations/new.html.erb

and will also need to store info in the db - we'll also need to give the user the option to change in the user edit screen, but that can be broken out as a separate task ...

praveendhawan:
We want a default unchecked checkbox (stored as a boolean in the user model) with wording like: "I give permission for AV to send me occasional emails"

PR #2298, opened 2018-04-18
[WIP] Add support for GDPR Compilance

praveendhawan:
This will provide the user with the functionality to opt-in to the newsletter at sign-up.

FIxes - [2216](https://github.com/AgileVentures/WebsiteOne/issues/2216)


tansaku:
thanks for starting on this @praveendhawan - interesting that your change makes so many updates to the  schema.rb file - I wonder if we need to standardize on something to avoid what seem like platform specific (?) changes?

app/controllers/application_controller.rb
```diff
@@ -21,6 +21,8 @@ def configure_permitted_parameters
       :password_confirmation, :current_password,
       :display_email, :display_profile, :display_hire_me,
       :receive_mailings, :status])
+    
+    devise_parameter_sanitizer.permit(:sign_up, keys: [:emails_opt_in])
```
tansaku: is this working now @praveendhawan ?

praveendhawan:
@tansaku ready for review

tansaku:
@praveendhawan this looks pretty good - does it affect any cucumber tests - should we prefer a cucumber test that nails down the user experience and tests end to end?

tansaku:
also I note we have an existing flag (`receive_mailings`) which I think the user can set by editing their profile - perhaps we should use this (maybe refactor the name) instead of adding a new boolean?

praveendhawan:
@tansaku One spec is failing, but I could not find its relation to the changes I have done. 
I will check and restart the build to see if that solves it, which would mean the spec is flaky. Also will run that particular spec on local to see If I can find out something.

Also, we can use the `receive_mailings` flag. Should I rename it to `opt_for_email_notifications`?

praveendhawan:
@tansaku I think that was due to branch not being updated with develop. I have pushed it after updating the branch with develop.

praveendhawan:
@tansaku just a question, how this opt-in for emails will work for social signups, like 'github' etc.

tansaku:
@praveendhawan I guess supporting social signups is another ticket - we'll need an intermediate page

@maryam21 also working on related functionality - let's refactor name of boolean after both of your PRs are merged

tansaku:
@praveendhawan can we remove the migration and ensure that we use the existing `receive_mailings` as the boolean

tansaku:
@praveendhawan if you can't get to this this week, please let us know and someone else will take over - thanks :-)

nisevi:
:rocket:

PR #2315, opened 2018-04-27
RSVPping AV events

### LINKED ISSUE #1957: RSVPing AV events
Not only does it encourage attendance, but it also informs the host if they need to cancel the event.

### LINKED ISSUE #2309: Add a 1-click [Postpone] button for events
_**As an** event host or attendee_ 
 
_**So that** I can easily postpone the next scheduled meeting, without needing to edit event details and alter the original start date of the event_
 
_**I would like to** confirm the postponement of an event, after easily indicating that it has been postponed on the event details page_

maryam21:
#### Description

Enable event creator to specify that they cannot host the event.
#### Motivation and Context

Fixes #1957 
Fixes #2309
#### Types of changes

#### Checklist:


app/controllers/events_controller.rb
```diff
@@ -65,6 +65,11 @@ def destroy
     redirect_to events_path
   end
 
+  def event_attendance
```
tansaku: if we're being RESTful we might prefer not to create new actions in the controller and instead prefer to create a new attendance object, but you've got the RESTful route set up, so this seems a good intermediate if we can set up the refactoring/feature ticket that would pull attendance out into a different model ...

tansaku:
great work - thanks @maryam21 

I'm just thinking that perhaps we should be updating event attendance via the existing update method? or would that be too complex?

Also, might we call the boolean "creator_will_attend" or something that makes it clearer what it means?

What do you think?

app/models/event.rb
```diff
@@ -266,6 +266,14 @@ def modifier
     User.find modifier_id
   end
 
+  def event_attendance
```
tansaku: is this method still being used? if not, perhaps we can delete it?

features/events/rsvping_event.feature
```diff
@@ -0,0 +1,55 @@
...
+    Given the following users exist
+      | first_name | last_name | id  | password  | email                  |
+      | Alice      | Jones     | 401 | 12345678  | alicejones@hotmail.com |
+      | John       | Doe       |     | password  | john@doe.com           |
+    And following events exist:
+      | name       | description             | category        | creator_id | start_datetime          | duration | repeats | time_zone | project | repeats_weekly_each_days_of_the_week_mask | repeats_every_n_weeks | creator_attendance |
+      | ClientMtg  | Weekly client meeting   | ClientMeeting   | 401        | 2018/07/19 11:00:00 UTC | 150      | never   | UTC       |         |                                           |                       | true       |
+      | Meeting    | Weekly meeting          | Meeting         | 401        | 2018/07/21 11:00:00 UTC | 150      | weekly  | UTC       |         | 31                                        | 1                     | false      |
+
+  Scenario: Non-logged in users should not see Cannot Attend button
```
tansaku: we might be able to dry these out with Scenario Outlines?  but no rush - could be a refactoring ticket so we can get this deployed ...

tansaku:
Great stuff @maryam21 - I opened a refactoring ticket for the cuke: https://github.com/AgileVentures/WebsiteOne/issues/2507

I also just noticed that there are a load of tech stack related things in the schema.rb - I wonder if that will cause problems?  Maybe you could refresh the schema.rb? or just remove from the commits?

maryam21:
Thanks for the review. I just clean up the shema, I don't know how that stuff got there.

tansaku:
wonderful! thanks @maryam21 :-)

Issue #2345, opened 2018-05-04
Remove chatlio from websiteone


nisevi:
`chatlio` should be removed since is no longer being used, but the javascript library is still being delivered to the client.

![image](https://user-images.githubusercontent.com/3024789/39627504-6fe52578-4fa6-11e8-8765-ff7bbbac3ce5.png)

Things to remove:
- Reference to the library: //= require chatlio
- File: app/assets/javascripts/chatlio.js

PR #2346, opened 2018-05-04
Remove chatlio from websiteone.

nisevi:
Fixes https://github.com/AgileVentures/WebsiteOne/issues/2345


nisevi:
@tansaku only two files are related to code. All the other files are related to this path `WebsiteOne/features/support/fixtures/req_cache/` and since this functionality is no longer present in the project I considered unnecessary to keep more than 200 files around for cache-ing something that is never used :+1:

tansaku:
thanks @nisevi :-)

PR #2376, opened 2018-05-15
Set tests to fail on first_try

NickSchimek:
#### Description

When tests fail on CI there showing as passing. We used to run a
second_try, but that was removed for some reason. Setting first_try to
fail so that failing tests are shown as failing.

#### Motivation and Context

Created this change to see if it affects CI reporting of failing tests.


nisevi:
@NickSchimek @tansaku I'm merging this in, I need this change :+1:

PR #2380, opened 2018-05-15
Control letter_opener with env variable

maryam21:
#### Description

Add environment variable to enable letter opener.
#### Types of changes

#### Checklist:


tansaku:
thanks for this @maryam21 - forgive me if I've forgotten, but why did we want an ENV var for the letter_opener? 

Would it make sense to always have it enabled in development mode?

maryam21:
When I was testing in my local env it did get in the way because I was using c9 instance and whenever an email gets sent it would give an error because there's no browser...And also because when you were working on GDPR stuff you suspected that letter_opener might have caused the errors and that's why I did it in a separate PR.

tansaku:
ah, c9 - hmm, but might be good to have it on by default and then enable switching it off in c9 - wonder if we can autodetect presence of C9

Also turns out it was not the cause of those GDPR issues ...

maryam21:
I am not sure about detecting c9 I will try looking at that, but I agree it would be better to have it by default.

config/environments/development.rb
```diff
@@ -25,7 +25,7 @@
   config.assets.debug = false
   config.assets.compress = false
 
-  config.action_mailer.delivery_method = :letter_opener
+  config.action_mailer.delivery_method = :letter_opener if ENV['LETTER_OPENER'] == 'true'
```
tansaku: how about:

```
config.action_mailer.delivery_method = :letter_opener unless ENV['LETTER_OPENER'] == 'false'
```

maryam21: oh yeah you said it would be easier in production that way. I will change it.

tansaku:
great @maryam21 - so I just set LETTER_OPENER to false on production - and now it will be used elsewhere unless we turn it off - I think that's the best setup :-)

PR #2383, opened 2018-05-15
Make interceptor work in all environments

maryam21:
#### Types of changes

#### Checklist:


tansaku:
thanks for this @maryam21 - we might even want to think about what we want the default behaviour to be - does it make more sense to have interceptors in place by default, and create an ALLOW_EMAILS flag that turns them off?

maryam21:
Wouldn't having the interceptor by default impact production environment because emails won't be sent unless we explicitly set ALLOW_EMAILS.

tansaku:
@maryam21 yes, but we can explicitly enable that setting in production and then it will always work there 

note that production is the only place where we want emails going out, so we want the default on any other server for them to be intercepted, and for local machines developers should all have letter-opener and then interceptor just in case ...

maryam21:
yes makes sense. I will change it that way

PR #2384, opened 2018-05-15
Activate second try.

nisevi:
Activate cucumber second try.

### DIFF
--- a/lib/tasks/ci.rake
+++ b/lib/tasks/ci.rake
@@ -8,6 +8,6 @@ unless Rails.env.production?
 
   namespace :ci do
     desc 'Run Rspec and Cucumber'
-    task tests: [:spec, 'cucumber:first_try']
+    task tests: [:spec, 'cucumber:first_try', 'cucumber:second_try']
   end
 end
--- a/.simplecov
+++ b/.simplecov
@@ -8,4 +8,4 @@ SimpleCov.start 'rails' do
   add_group 'Services', 'app/services'
   add_group 'Mailers', 'app/mailers'
   add_group 'Jobs', 'app/jobs'
-end
+end if ENV['COVERAGE']
--- a/config/cucumber.yml
+++ b/config/cucumber.yml
@@ -7,5 +7,5 @@ default: <%= std_opts %> features
 ci: --format pretty --strict features
 wip: --tags @wip:3 --wip features
 rerun: <%= rerun_opts %> --format rerun --out rerun.txt --strict --tags ~@wip
-first_try:  NEVER_FAIL=false  <%= std_opts %> --format rerun --out rerun.txt features --strict --tags ~@wip
+first_try:  NEVER_FAIL=true  <%= std_opts %> --format rerun --out rerun.txt features --strict --tags ~@wip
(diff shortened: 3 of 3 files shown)


nisevi:
@tansaku problems with coverage because of `second_try` ^^^

lib/tasks/ci.rake
```diff
@@ -8,6 +8,6 @@ unless Rails.env.production?
 
   namespace :ci do
     desc 'Run Rspec and Cucumber'
-    task tests: [:spec, 'cucumber:first_try']
+    task tests: [:spec, 'cucumber:first_try', 'cucumber:second_try']
```
NickSchimek: second_try only runs the failing tests which is why the coverage is failing. If we re-run first_try 
```
task tests: [:spec, 'cucumber:first_try', 'cucumber:first_try']
```
Then the coverage will be fine, but then we do spend another 15 minutes running all the tests...

NickSchimek: If that is something that would want to be done then we can copy the code from first_try over to second_try since first_try is set to never_fail.

nisevi: I would prefer to find a way to tell `simplecov` to stop collecting data from tests and use what it has when enters `second_try`. So that way we wouldn't have to wait another 15 minutes just because we don't know why test are failing.

In case there is no way to do that, what you propose is a good patch for the moment.

The optimum approach it will be to not use `second try` at all and fix those intermittently failing tests, but we can create an story for that.

NickSchimek: I agree we should not even have to use a second try. I think it may be worth to hold off on implementing a second try and collect data about how often we are getting these intermittent failures and on what tests. If it happens rarely than it may not be worth spending any time on it, but if it does happen often than we will have the data to narrow in on the failing tests and figure out what is possibly going wrong.

tansaku: @NickSchimek it still regularly happens and blocks the websiteone deployment process - it's a real drag - we have some tags to exclude some tests (or was that localsupport?) and I think semaphore gives us stats on failing tests and execution time ...

https://semaphoreci.com/agileventures/websiteone/insights/test_files/698

It's also the case that it's really difficult to find the error output when a single test fails in first_try - having second_try run the tests again means it's then easy to see the output.  I'd give a lot of money if the failing CI would post the actual error message in the PR, but I don't know any CI that does that.  I believe there is a flag for cucumber now that will display only details for the failing test:

https://github.com/cucumber/cucumber-ruby/issues/1094#issuecomment-312245006

that we could use to achieve the same effect - I'd love not to have the second_try, but the failing tests double or even triple the time it takes us to deploy code since an intermittent fail prevents a deploy at some point in the pipeline and we have to move from develop to staging to production.

I would have thought we could just disable simplecov for the second_try somehow? @nisevi you said you opened a ticket with someone? can you post a link to the issue?

NickSchimek: @tansaku @nisevi I read the simplecov docs and there is a way to disable it for the second try. I will submit the necessary changes.

tansaku:
wonderful - thanks @NickSchimek

lib/tasks/ci.rake
nisevi: Awesome @NickSchimek !! thank you for taking a look at this :+1:

PR #2387, opened 2018-05-16
Remove pg stats

NickSchimek:
#### Description

PG stats are enabled on Heroku postgres databases by default. Enabling
PG stats uses up more memory and are not needed on local dev machine
since PG stats are used to collect production statistical data that can
used for optimization plans.


tansaku:
thanks @NickSchimek - shall we go the whole hog and remove it and add the file to .gitignore?

tansaku:
I'll merge this and we can do another PR for removing if we all agree ...

NickSchimek:
I'd prefer to leave the schema checked in.

PR #2388, opened 2018-05-16
[WIP] worked out how to pass hard coded user params along with omniauth ...

### LINKED ISSUE #2385: GDPR options for Github and Google Signups
At the moment the Github and Google Signups will ignore the email consent options.  They will default to false so that's okay in terms of not getting reported, but we'd like people who ask to receive emails (and sign up via a third party) to get those emails ...

tansaku:
fixes #2385

### DIFF
--- a/app/controllers/authentications_controller.rb
+++ b/app/controllers/authentications_controller.rb
@@ -5,6 +5,7 @@ class AuthenticationsController < ApplicationController
 
   def create
     omniauth = request.env['omniauth.auth']
+    omniauth_params = request.env['omniauth.params']
     authentication = Authentication.find_by_provider_and_uid(omniauth['provider'], omniauth['uid'])
 
     @path = request.env['omniauth.origin'] || root_path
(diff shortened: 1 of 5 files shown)


mattwr18:
Failing this test:
`rspec ./spec/controllers/authentications_controller_spec.rb:87`
with this message:
```
NoMethodError:
       undefined method `merge' for nil:NilClass
```

mattwr18:
@tansaku there is an open issue on omniauth related to this:
https://github.com/omniauth/omniauth/issues/937
Actually, there seem to be a lot of open issues related to testing its behaviour :(
https://github.com/omniauth/omniauth/issues/932
With no response from them... 
https://github.com/omniauth/omniauth/issues/934
There is this wiki, which I wonder if you have seen, but nothing about params particularly:
https://github.com/omniauth/omniauth/wiki/Integration-Testing
_wonder if you might get me caught up to what you have already tried_

mattwr18:
Ahh, right so manually testing it seems like in fact the params do persist to the database, was that your experience @tansaku?

tansaku:
@mattwr18 can't remember :-(  my vague sense is that this was "working" but couldn't get a way to test it automatically ...

PR #2417, opened 2018-06-08
Decrease running time on Cucumber tests

### LINKED ISSUE #1496: investigate why these four cukes take so long
1 features/events/edit_past_event.feature		
2 features/events/list_repeating_events.feature	
3 features/hookups/hookups_pending.feature		
4 features/hangouts/edit_hangout_url.feature

according to semaphore each is taking longer than 40 seconds on average

from https://semaphoreci.com/agileventures/websiteone/insights

NickSchimek:
Multiple tests rely on a user being signed in. When a user needs to be signed in for a test then prior to that test run capybara is driving to the webpage, filling out the form, submitting, and waits for a reply.

Having a method that stubs the sign in process will not only improve the tests listed below, but all tests that can run with a stubbed sign in.

Fortunately Devise has a Module called Warden that does this. Configured Warden and set most tests to sign in via Warden.

Also removed all capybara methods and assertions in the Cucumber background job since it is a setup job that doesn't/shouldn't test functionality as it should be used to setup a state for the tests to run in. Since the background job is ran before each test in the feature than improving this one job has improved overall test speed.

Fixes #1496 

#### Description

Semaphore has listed the following tests as taking a long time to run.

- [x] 1 features/events/edit_past_event.feature
- [x] 2 features/events/list_repeating_events.feature
- [x] 3 features/hookups/hookups_pending.feature
- [x] 4 features/hangouts/edit_hangout_url.feature

Will also update other tests where possible to use Warden's login_as method
- [x] Update tests across site to use Warden's stubbed login method

#### Motivation and Context

To decrease overall test run time

Overall the whole test suites runs about a minute faster than before. 
The edit_past_event.feature runs about 2x faster now
And not much improvement on the other tests as there is only 1 or two scenario's there and they already run in about 1 - 2 seconds.

#### Checklist:

### DIFF
--- a/features/article_vote.feature
+++ b/features/article_vote.feature
@@ -31,7 +31,7 @@ Feature:
 # Logged in
 
   Scenario: I should see the links to Up/Down Vote an article on the article show page
-    Given I am logged in
+    Given I have logged in
     And I am on the "Show" page for article "Ruby is on Fire"
     Then I should see link "Up Vote"
     And I should see link "Down Vote"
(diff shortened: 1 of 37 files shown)


NickSchimek:
@tansaku sorry should've explicitly mentioned that it's tested in the users/sign_in.feature & users/sign_off.feature

PR #2444, opened 2018-06-22
Remove references to jetbrains

### LINKED ISSUE #2395: remove jetbrains sponsorship logos
remove jetbrains sponsorship logos since we are no longer sponsored by them.

Use Github search to see all the places we refer to them in the codebase:

https://github.com/AgileVentures/WebsiteOne/search?utf8=%E2%9C%93&q=jetbrains&type=

mattwr18:
No longer sponsoring AV
Closes #2395


.gitignore
```diff
@@ -15,6 +15,10 @@
 /db/*.sqlite3
 /db/*.sqlite3-journal
 
+# Ignore database.yml file
+
+/config/database.yml
```
nisevi: @mattwr18 I can help you to get the configuration you need in your computer for using this `database.yml` there is no need to remove it :+1: 

Alternatively you can take a look here https://gist.github.com/nisevi/49da4b76da393b9deaeafa0759334c6c :robot:

mattwr18: I added it to the ` .gitignore` because my password is different. I have `postgres` working and am able to access the database. If there is a preferred way of doing this, I can change it. 
I know with RFM there is a `database.template.yml` so that people can copy it to a `database.yml` and use their credentials.

nisevi: I agree that you can have a different configuration on your computer but that is not enough to make a change that will impact on the configuration of all developers contributing to this project, also in a PR that it has nothing to do with the story on Pivotal Tracker.

To be more polite about the change that you are trying to make, you could create another story and discuss about it in the Slack channel. In this way all devs will have the opportunity to make their statements :+1: .

mattwr18: Sure. I didn't make this change without first talking to @tansaku. I'm happy to change it though.

tansaku:
looking great @mattwr18 

one thing I note is that the package.lock is updated - not sure if we necessarily want that as part of this ...

Also those premium erb html files are actually all redundant now that we have them in the AgileVentures/AgileVentures repo so we might want to open another ticket to have them removed to avoid confusion

tansaku:
@mattwr18 do you need help removing the package-lock.json from this PR?

mattwr18:
Hey @tansaku, sorry I just saw your messages now. Maybe, I do need a hand. I ran this command `git checkout upstream develop -- package-lock.json` to revert to the `package-lock.json`, but then it still appears in the `PR`, I think. Is there a better way to do this?

Should I open an issue on `Waffle` to remove the premium erb html files?

mattwr18:
Actually, it seems like that worked. Shall I do the same for the `database.yml` file since there aren't any actual changes, just a trailing whitespace being detected?

tansaku:
@mattwr18 that's fine - great work - will get this merged in :-)

PR #2491, opened 2018-07-24
[WIP] Acceptance test for private event feature

### LINKED ISSUE #1596: Private events for premium members

tansaku:
fixes #1596


tansaku:
great progress on this @mattwr18 

although it's getting difficult to follow as we're ending up with a lot of unrelated changes in the PR -- not sure the best way to fix that - might be to cherry pick into a new one ... maybe ...?

tansaku:
hmm, although maybe this PR will just get closed when we merge #2579 ...

mattwr18:
Hey @tansaku, so yeah I was trying to work out how to use this PR when collaborating at the English Hack Night and all of them have rights to the RubyFloripa forked repo and so we pushed to that and put in a PR and that made this one redundant...

PR #2498, opened 2018-07-28
Add Slack channel link to the projects show page

### LINKED ISSUE #770: get mention of which slack chat room to go to on every project page
Seems like people don't generally know how to get to the right chat rooms for a project

NickSchimek:
#### Description

Added a link so that users can navigate to the projects slack channel

#### Motivation and Context

This is so that individuals can easily navigate to a projects slack channel.

Fixes #770 

Added tests to ensure link is present.
screenshot 2018-07-28 09 59 53

#### Types of changes

#### Checklist:

### DIFF
--- a/features/projects/connections.feature
+++ b/features/projects/connections.feature
@@ -9,8 +9,8 @@ Feature: Projects should show links to the connected APIs
       | Bill       | Bob       | Bill@example.com | true              |
 
     Given the following projects exist:
-      | title | description | status   | author | pivotaltracker_url                               | github_url               |
-      | hello | earthlings  | active   | Bill   | https://www.pivotaltracker.com/n/projects/742821 | https://github.com/hello |
+      | title | description | status   | author | pivotaltracker_url                               | github_url               | slack_channel_name |
+      | hello | earthlings  | active   | Bill   | https://www.pivotaltracker.com/n/projects/742821 | https://github.com/hello | hello_earthlings   |
 
   Scenario: I can click a link to the GitHub project page
     When I go to the "hello" project "show" page
--- a/features/projects/connections.feature
+++ b/features/projects/connections.feature
@@ -11,17 +11,27 @@ Feature: Projects should show links to the connected APIs
     Given the following projects exist:
       | title   | description | status   | author | pivotaltracker_url                               | github_url               | slack_channel_name |
       | hello   | earthlings  | active   | Bill   | https://www.pivotaltracker.com/n/projects/742821 | https://github.com/hello | hello_earthlings   |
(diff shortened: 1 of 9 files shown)


NickSchimek:
This was implemented under the assumption that slack channels had the same name as the corresponding project, which I just realized that there are some slack channels with slightly different names than the project title. 

One such project is the `asyncvoter` project (Well it's AsyncVoter, but slack doesn't care about case) where the slack channel is `async_voter`. This will cause a 404 page not found error.

I personally feel that the slack channel names should match the project titles, Is there any opposition to not keeping the names the same across `both slack and websiteone`?

hmm, I suppose the main constraint is that project titles are allowed to have spaces, but slack channel names cannot. Looking at the Asyncvoter's field data I don't see anything with the name `async_voter` that could be used for the slack channel. I remember you suggesting to use the Channels Hash to get the corresponding slack channel, but I noticed not all of the Projects are on that list.

I suppose I could build a drop down box where the User can choose the slack channel to go to, but if the initial concern was that User's didn't know which slack channel to join for the project than I don't see how having a list of channels is any better than trying to find the channel in slack.

Do you think it makes sense to introduce a new field called `slack channel name` or `chat channel`?

NickSchimek:
@tansaku I decided to add the new column. Please let me know if you visualize this being done a different way. While adding tests for the slack channel I also added additional tests for the github and issue tracker connections in doing so I uncovered a bug that was occurring only in the test environment. I patched the bug in the following commits:

f41f423d Update test to check the views else route
94f8a76d Remove building source repository
28c0e6a8 Update tests now that source repo is not built

If this PR is merged would you prefer I create a script to run in order to update the current projects with their slack channel names or notify the project owners to update their Projects with the slack channel name?

features/projects/connections.feature
```diff
@@ -0,0 +1,44 @@
...
+    Given the following users exist
+      | first_name | last_name | email            | receive_mailings  |
+      | Bill       | Bob       | Bill@example.com | true              |
+
+    Given the following projects exist:
+      | title   | description | status   | author | pivotaltracker_url                               | github_url               | slack_channel_name |
+      | hello   | earthlings  | active   | Bill   | https://www.pivotaltracker.com/n/projects/742821 | https://github.com/hello | hello_earthlings   |
+      | Bat Man | All bat     | active   | Bill   |                                                  |                          |                    |
+
+  Scenario: I can see a link to the GitHub project page
```
tansaku: I wonder if some of this could be condensed down with a "Scenario Outline"?

NickSchimek: Good point. I will take a look at that!

tansaku: feel free to open a new refactoring ticket for that :-)

PR #2524, opened 2018-08-07
[WIP] 797 projects index page redesign

### LINKED ISSUE #797: Projects index page redesign
Imported from https://www.pivotaltracker.com/story/show/79660056

**Story**
As an agile ventures user
So that I can (find a new | create a new | find my | promote my) project(s)
I would like to see a fully featured projects index view

**Supporting documents**
Feature request on AG: http://agileventures.org/projects/websiteone/documents/project-index-page-ux-proposal
Direct link to wireframes: https://drive.google.com/file/d/0B9FqKjpP8_MUQXBlcWZ2R3JzV2c/edit?usp=sharing

**Tasks:**
- No tab for current users projects
- No search
- No featured project

mattwr18:
Search for projects by tech stack

fixes #797


NickSchimek:
hey @mattwr18 - I spun up the branch and the drop down to search for tech stacks is blank. I went to add technology stack to a project, but didn't see where I could add any. Do you have any screen shots of your implementation that you can share?

Also when I ran the migration I got this schema diff. Do you know why?
screenshot 2018-08-07 16 11 56

mattwr18:
I added  this migration: 
screenshot at 2018-08-07 22 09 11

Can you run `rake db:seed`? That should add the tech stacks to the projects. 
Although, ultimately maybe this information can be requested from `GitHub`?

mattwr18:
So, that other migration I removed because of a few conversations with developers telling me it's not good practice to have a column that is an array of data.

tansaku:
looking good @mattwr18 :-)

NickSchimek:
I seeded the db. I see the drop down and can search now. That is pretty cool!

tansaku:
only thing I'm thinking is whether we should call the tech "technologies" rather than "stacks"?  I think github just lists technologies and we might want to allow for, say, "Ruby on Rails" project to add React later on and still be searched for ...

although I guess projects can be associated with multiple stacks?

Just wondering if the terminology of "stack" makes sense?   I sort of think of a stack as comprising a few things, e.g. "Ruby on Rails" includes Ruby, so when searching for Ruby I'd expect all the Rails projects to show up, but not necessarily expect the project owner to mark it separately as Ruby and Ruby on Rails ...?

Do I make any sense? ignore me if not :-)

tansaku:
yeah, certainly we'll pull the info in from github, but this is the first step, and even if/when we're pulling it in from github we'll store the data in these tables.

was just thinking it might me comprehensible/make-sense for folks in future if we refactored to use the word "technologies" rather than "stacks" but that could easily be another refactoring ticket.

I was just thinking about the domain model - see https://domainlanguage.com/ddd/

either way, not worth burning too much brain sugar on it :-) make the change now, make a refactoring ticket, or not bother - all are good :-)

tansaku:
thanks @mattwr18 great work here :-)

I've been running locally and I can see the infinite scroll working - great job!

Here's what it looks like for me:

tansaku:
As regards the formatting/coloring I wonder if it would make sense to follow the pattern we have on the users page where we don't have a highlighted background for the items, but only highlight on mouse-over?  Just a thought.

Also, when I go to the users page I think something weird is happening with infinite scroll there - check it out - maybe something is bleeding over from the projects?

I'd much prefer infinite scroll on the users too, but that's another story :-)

tansaku:
I wonder if there's a risk of trying to get too many new features in at once here?  Would it make sense to separate out the following:

1) infinite scroll
2) change in formatting 
3) search by language

And see those as three separate features that might be worked on independently, or at least submitted as three separate PRs that could be evaluated independently and even deployed to production independently?

The counter-argument might be that we want to see how they all work in combination (which we probably do) but it might be a much bigger job to get all three of them done together smoothly, and it might be easier to split them into three separate things?

mattwr18:
Ok @tansaku, I've put in those three `PR`s

tansaku:
thanks @mattwr18 - should we then close this PR?

mattwr18:
I guess it would be sensible to do that @tansaku, although I'll always have fond memories of this one. :P

PR #2526, opened 2018-08-07
Better dev flow with docker

### LINKED ISSUE #2525: Cannot edit source files without rebuilding entire docker containers
So if i edit a simple view in the source files, i cannot see that reflected in the container and app by running docker compose up --build and taking advantage of caching.  I have to do docker/setup.sh which is really slow

mtc2013:
Fixes #2525

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ COPY scripts /WebsiteOne/scripts
 RUN dos2unix scripts/copy_javascript_dependencies.js
 
 RUN npm install --unsafe-perm
+RUN npm install -g phantomjs-prebuilt --unsafe-perm
 
 COPY . /WebsiteOne
 COPY ./docker/config/database.yml ./config/database.yml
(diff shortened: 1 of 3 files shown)


mtc2013:
@tansaku please review

tansaku:
@mtc2013 looks fine to me, but I've not tested it  - @mattwr18 any thoughts?

I wonder can the docker be used by semaphore so that it gets tested as part of every build?

mattwr18:
yes, these changes we were working on together @tansaku ... the old configuration was lacking in that it didn't allow for developers to change files and have those reflected on `docker-compose up --build`

Although, I had checked this, I hadn't recently and @mtc2013 brought it to my attention. 
Also, he added a test database and phantomjs install in the Dockerfile. 

There seems to be support for that @tansaku, I found this:
https://semaphoreci.com/docs/docker.html
I'll have to have a closer look though.

mtc2013:
@tansaku @mattwr18 in principle testing for ci in docker might be desirable, but we will have to deal with issues i reported here:

https://github.com/AgileVentures/WebsiteOne/issues/2527

mtc2013:
@tansaku we are ready to pull this in, yes?

tansaku:
yup all done

PR #2539, opened 2018-08-13
Search by programming language

### LINKED ISSUE #2538: Search for projects by language(technology)
Allow user to be able to search for projects based on the technology stack used.
This info can be pulled in from GitHub and added to the database.

mattwr18:
Set up search by programming language functionality, tests, rake task
Closes #2538 
#### Description
Add simple forms for search field and submit button for projects
Add populate languages dropdown method for projects
Add filter projects method that when there are project params calls a search by language model method
Add search by language model method that returns projects with same languages as search params
Add js erb file to receive the filtered projects and replaces the entire project list paginated
#### Motivation and Context
This allows for users to more quickly identify project that use programming languages they already know quite well and/or those they would like to learn and the projects that they can contribute to.

Use Cucumber scenarios to mimic user interaction
#### Types of changes

#### Checklist:


app/controllers/projects_controller.rb
```diff
@@ -79,6 +87,18 @@ def unfollow
     flash[:notice] = "You are no longer a member of #{@project.title}."
   end
 
+  def populate_languages_dropdown
```
tansaku: should this and the method below be in the private zone perhaps?

lib/tasks/scheduler.rake
```diff
@@ -8,6 +8,10 @@ task :fetch_github_last_updates => :environment do
   GithubLastUpdatesJob.run
 end
 
+task :fetch_github_languages => :environment do
```
tansaku: wonderful!  do we have a devops test covering it?

app/jobs/github_languages_job.rb
```diff
@@ -0,0 +1,21 @@
...
+        languages.keys.each { |language| p.languages.create(name: language) }
+      rescue StandardError => e
+        ErrorLoggingService.new(e).log("Updating the languages for #{p.github_url} may have caused the issue!")
+      end
+    end
+  end
+
+  private
+  def client
+    @client ||= Octokit::Client.new(:access_token => Settings.github.auth_token)
```
tansaku: don't think our production.yml has Settings.github.auth_token set ...

tansaku:
I'm getting this error now I sorted the github tokens:

```
GET https://api.github.com/repos/visualizer/AgileVentures/languages: 404 - Not Found // See: https://developer.github.com/v3/repos/#list-languages
```
```
Updating the languages for https://github.com/saasbook/hw-rails-intro/blob/master/docs/part_1.md may have caused the issue!
GET https://api.github.com/repos/hw-rails-intro/saasbook/languages: 404 - Not Found // See: https://developer.github.com/v3/repos/#list-languages
```
same for all of them I think ...

mattwr18:
hey @tansaku, thanks for the review... So, I spent the majority of the day working on this, took me a while to get the test passing and made some modifications to the GithubLanguagesJob. 

I also made those methods private, had a pairing session with @aonomike where I walked him through what we are trying to do with this ticket. I noticed a couple of things I am not so happy with and worked on debugging them, but stopped until I can get some clarification from you. 

I noticed that the pagination doesn't work properly for filtered projects, but just goes to the second page of projects because it doesn't hit the `index.js.erb` file again. Also, the `filter_projects` method call in the `index` method of `ProjectsController` gets called again cause there are still `params[:project]`. 

I had a go at calling the `filter_projects` method directly from the `simple_form_for` dropdown, which has the benefit of avoiding JavaScript altogether, but means a `route`, and some non-DRY code needs to be placed in a `filter-project.html.erb` file.

Also, with the `infiniteScroll` functionality, it seems to work for filtered results as well. I think there would still be an issue with the pagination if JS were not available for some reason though.

tansaku:
great work here @mattwr18 - I'm starting a more detailed review - the github task ran locally and I got lots of languages - yay!

tansaku:
I did notice that quite a few projects still didn't get coded due to rate limiting - we'll see how it goes on production (I have a clone of production locally) but makes me think we might need to restrict to coding up active projects or potentially start looking at throttling the process or other ways of backgrounding - but that can be a separate ticket/feature/issue

tansaku:
thanks @mattwr18 - I think I have it working with pagination given a few tweaks of the code:

```
    @projects = Project.order('status ASC')
                       .order('last_github_update DESC NULLS LAST')
                       .order('commit_count DESC NULLS LAST')
                       .includes(:user)
    if params[:project]
      @language = params[:project][:languages]
      @projects = @projects.search_by_language(@language, params[:page])
    else    
      @projects = @projects.search(params[:search], params[:page])
    end
```

and _filtered_project_list_html.erb set to

```
<% @projects.each do |project| %>
  <%= render 'listing', project: project %>
<% end %>
```

mattwr18:
Really!? It was that easy!? HA! I'm going to bed!! hehe just kidding, that's great if it's working! Also, rubocop will be happy about breaking that @projects line down into a few. gotta say that was bugging me a little bit :stuck_out_tongue:

app/controllers/projects_controller.rb
```diff
@@ -106,7 +115,22 @@ def get_current_stories
...
+    @projects_languages_array = Array.new
+    Language.all.each do |language|
+      @projects_languages_array << language.name
+    end
+  end
+
+  def filter_projects
+    @language = params[:project][:languages]
+    @filtered_projects = Project.search_by_language(@language, params[:page])
```
tansaku: I think the rule of thumb here is to avoid starting a new query with a `Project.` if we can chain on to the existing query and vary it according to circumstances that will generally be smoother - although now always :-)

mattwr18: Are you suggesting that I change this @tansaku or should I go with the implementation you used :arrow_up: and make sure it works for me and passes all the tests?

tansaku: @mattwr18 - up to you - I think I fixed the issue with the pagination working across the search by language (although I might be fooling myself) - and you should totally feel free to use my code/fix.

but if you want to re-implement, or use part of it, or use it and re-factor that's all good too.  Depends on your energy level and what's best for your learning ...

I guess technically we should have a test that shows search by language working across the pagination - although I can see how that might seem redundant if we're going to replace with infinite scroll soon - although of course at that time the test should evolve into checking that form of pagination ...

Am I making any sense? :-) decision in your hands of course :-)

mattwr18: Judging from your emoji response, and re-reading my comment, I think I didn't express what I was trying to ask. I was a little unclear how to proceed with getting this in, and briefly thought that maybe you would just commit your changes. 

I think it is better for me to do it though. There are still a couple things I don't like though. 
   ~~1. It seems like I needed to add a line `@project = Project.new` to the index page to get the 
       languages dropdown to work. It seems you didn't need to do this~~ Actually, just figures this 
       out.
       2. The other is that the index method is now 17 lines!! 4 lines are to set the initial @projects 
       which could be refactored into a private method in the ProjectsContoller? 4 lines are the 
       respond_to block that I'm not aware of a way to refactor.  Should the refactoring be a 
       separate ticket?

I have added test coverage for search by language working across the pagination.

tansaku:
hi @mattwr18 so I was reading a comment by you here indicating that maybe I'd misunderstood something and then I refreshed the page and the comment went away and maybe that's related to the comment being attached to some review thread on some code that now got moved out - so either which way my sincere apologies for confusion arising :-)

I'm looking forward to getting back to the UK so we can resolve these things in scrums rather then pure text :-)

Anyway, I think you were asking about refactoring the index method, which is long - I kind of purposely unpacked it with the intention of playing around with it to understand it, but having got it working we can totally refactor it - pull out new methods and that can be done in this PR or be a refactoring ticket - it's all good.

Maybe I'm being too flexible :-) Anyway, right now I'd lean towards saying the priority is to add a test that checks that pagination works correctly with language search.  Test coverage trumps refactoring in my book - confidence in refactoring success goes up as tests are greener and have wider coverage.

Sorry I woke up late today so our chance to discuss this in real time was limited.  Looking forward to getting this merged in, but no rush on my side - you're spinning lots of plates in lots of projects (and being a huge dynamo of enthusiasm and good feeling for the whole community - thanks so much!) - hope the stacatto text communication isn't getting too irritating :-) only another 10 days or so ...

tansaku:
I think I'll merge this in and open a refactoring ticket and a ticket for maintaining the state of the simple form

app/jobs/github_languages_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def run
+    Project.with_github_url.each do |p|
+      begin
+        languages = client.languages("#{p.github_repo_name}/#{p.github_repo_user_name}")
+        language_names = languages.to_hash.keys
+        existing_languages = p.languages.collect(&:name).collect(&:to_sym)
+        incoming_languages = language_names - existing_languages
+        incoming_languages.each { |language| p.languages.build(name: language) }
```
tansaku: it looks like this should only create languages once in the Language table ... - in my local system I have many repeats - maybe from accidentally re-running the job, but I wonder here if we should put an index on the language table name to ensure uniqueness, and possibly re-structure this code to use `Language.find_or_create_by name: language`

mattwr18:
hey @tansaku, I ran into the problem of your response 'disappearing' as well and actually it is just hidden. If you scroll up to the comment `tansaku reviewed 2 days ago` under the `view changes` button there is a link that says `show outdated` which opens up an accordion that shows our  back and forth, but much more difficult to find, so let's talk on the main comment section :smile: 

I figured that is exactly what you had done, and I have a couple ideas on how to refactor it, although it won't cut down on many lines. 

So, you saw the test I added to, great! The only thing is that if you let me know which you will be merging in first I can modify the tests, so they don't cause each other to fail. 

I am not annoyed at all actually! Quite the opposite, I am really thankful that you are able to do so much and attend to so many people cause I can easily imagine a situation where I would be much more on my own trying to make decisions. That being said, I am definitely looking forward to having our normal routine back :smiley: 

As regards to the languages being unique, I agree and I wrote the GithubLanguagesJob to make sure that it doesn't add languages if they already exist for the project, but it doesn't check to see if that language exists in the Language table, which would be much better because then we don't need to have Ruby 2700 times in the database. I think this is what you mean correct?

I'll assume this is what you meant and refactor it accordingly/ make sure all the tests still pass.

app/controllers/projects_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
                        .order('last_github_update DESC NULLS LAST')
                        .order('commit_count DESC NULLS LAST')
                        .includes(:user)
-    @projects_languages_array = Language.all.collect(&:name)
+    @projects_languages_array = Language.all.collect(&:name).uniq
```
nisevi: Here you could do `Language.pluck(:name).uniq` 👍

mattwr18: hey @nisevi, Thanks for the tip!! :smiley:

tansaku:
thanks for this matt :-) glad we can find those old comments.

so I deleted all the languages and relations locally and re ran `rake fetch_github_languages` but it seems like multiple languages of the same name are still being created.  Maybe it's a mistake on my part, but perhaps we should explicitly [add an index](https://stackoverflow.com/a/15881833/316729) to the languages table to flush out the error, and perhaps adjust the devops test to explicitly check for the correct number of languages are being created?

apologies in advance if it's just a mistake on my side, but I've exceeded my github rate limit and can't easily test ... :-/

I was doing the following with a `rake db:migrate:redo`

```rb
class CreateLanguages < ActiveRecord::Migration[5.1]
  def change
    create_table :languages do |t|
      t.string :name, index:true
      t.timestamp
    end

    create_table :languages_projects, id: false do |t|
      t.belongs_to :project, index: true
      t.belongs_to :language, index: true
    end
  end
end
```

to add the index to the languages table without having to create a separate migration ... I tried re-running the test with the index, but it worked fine ... maybe it's all something weird on my side, but might be a good idea to have the test actually check that there is only one language created when we import two projects that both refer to it, just to make sure I'm not imagining things :-)

You might want to run the job locally yourself and check Language.all in the rails console - maybe you're doing that already?  I have a clone of the production data locally so maybe that's affecting things somehow ...

tansaku:
Ah and not just an index - might need a uniqueness flag - sorry for some reason I was thinking the index would enforce uniqueness -doh! :-)

tansaku:
Not sure what I'm doing wrong here - maybe need uniqueness validation on the Language model ...

mattwr18:
Hey @tansaku, I think I got it working and learned something in the process! So, the issue was that I was finding the language in the GithubLanguageJob, but then I was telling it to build the languages for each project by that language's name, creating duplicate languages!!!
What I learned is that I can find_or_create_by name and then append that Language object to the language project join table. Pretty awesome! Didn't know I could append an object to a database table!!
Also, I created rspec tests, and validations to ensure no duplicate languages are created. 
I don't have a copy of the production database, so see if it works for you! One thing I did was to add a `sleep 1.minute` to the rake task so that it would, in theory, not go over `Octokit`'s rate limit.

tansaku:
great work @mattwr18 :-)

app/models/project.rb
```diff
@@ -34,6 +36,13 @@ def self.search(search, page)
       .paginate(per_page: 5, page: page)
   end
 
+  def self.search_by_language(search, page)
+    includes(:languages)
+      .where("languages.name ILIKE ?", "%#{search}%")
```
tansaku: I think this partial match thing means that when we search for Java that we also match JavaScript - I think it might make more sense to just match the language name exactly ...

tansaku:
Thanks @mattwr18 - well done fixing the url update issue.

I think the extra table is just hanging around in an old schema of yours - you can probably delete the schema and re-create (running rake db:migrate:redo or something like that) and that extra table will go.

I think the place that you put the unique index won't quite have the effect that you intend. I'll add some comments in the code.

db/migrate/20180813125658_create_languages.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLanguages < ActiveRecord::Migration[5.1]
+  def change
+    create_table :languages do |t|
+      t.string :name
+      t.timestamp
+    end
+
+    create_table :languages_projects, id: false do |t|
+      t.belongs_to :project, index: true
+      t.belongs_to :language, index: true, unique: true
```
tansaku: I don't think the `unique: true` will have the desired effect here.

I think that we probably want the following:

```rb
class CreateLanguages < ActiveRecord::Migration[5.1]
  def change
    create_table :languages do |t|
      t.string :name, unique: true # <-- to ensure that language names are unique
      t.timestamp
    end
    create_table :languages_projects, id: false do |t|
      t.belongs_to :project, index: true
      t.belongs_to :language, index: true
    end
    add_index :languages_projects, [:project_id, :language_id], :unique => true # <-- to ensure that language project relations are unique
  end
end
```

app/models/project.rb
mattwr18: It took me a some research and playing around to get what you meant by partial match... so the % is a wildcard that matches partial words, as you pointed out. This is something I noticed that bothered me, but I hadn't fixed it. I have removed them and I think it works now.

db/migrate/20180813125658_create_languages.rb
mattwr18: Ok, I have changed it like above. The validations in the models will only rais an error if someone tries to create a language that is not unique, but these validations at the database level will ensure no one is able to append any languages which are not unique, is that a correct understanding of the difference/ need for both? Or should there be refactoring of the model validations?

mattwr18:
It's failing the build on the devops test, but it is passing locally. I'll have to have a look tomorrow, I guess.

mattwr18:
yeah, I saw... I remember getting this error a few times when the GithubLanguagesJob wasn't working correctly, but that was fixed. This devops test is passing locally. I am just running the whole suite of tests locally to see if it passes or not.

mattwr18:
Figured it out @tansaku, I had deleted some lines of code in the GithubLanguagesJob that check to see if the language exists in the join table before trying to find it to try to cut back on the complexity cause of that code smell. 

Interestingly, it was passes locally though.

tansaku:
great work @mattwr18 - ready to merge I think!

tansaku:
I'll also open a refactoring ticket for GithubLanguagesJob ... also we should check that all the other tickets I opened that you knocked off are closed :-)

PR #2540, opened 2018-08-13
Better dev part two

mtc2013:
This will make it so that the developer does not need to rebuild the image just because of an edit to basic source code.  However, changing dependencies, like gems, may still require a rebuild.


mtc2013:
@tansaku please start reviewing

diraulo:
I tried this setup and seems to be working fine. Made a few changes to views and didn't need to rebuild the image. Looks good IMO @mtc2013 but looks like you might need to update your branch with upstream?

tansaku:
@mattwr18 and friends seem very excited about this - will merge in

tansaku:
many thanks @mtc2013 :-)

Issue #2521, opened 2018-08-07
Slack icon not displaying


NickSchimek:
## Expected Behavior

The slack icon should display when a project is not associated to a slack channel. 
## Current Behavior

Currently the icon is a gear.
## Possible Solution

Update the fa icon in `app/views/projects/_connections.html.erb` line 26 to display the slack icon instead of the gear.

PR #2542, opened 2018-08-13
Add slack font awesome Icon

aonomike:
- Update the fa icon in app/views/projects/_connections.html.erb line 26 to fa-slack from fa-gear.

solves #2521


aonomike:
note @tansaku thanks

PR #2544, opened 2018-08-13
Add infinite scroll functionality, change tests

### LINKED ISSUE #2541: Favor infinite scroll over pagination

mattwr18:
Pagination works for users with javascript disabled
Closes #2541 
#### Description
Add javascript function in application.js
Call this function from projects, users, event_instances javascript files with params to support search functionality for projects and users
Fix tests 
#### Motivation and Context
Infinite Scroll is nice from a User Experience perspective

Tested with Cucumber/Capybara to scroll to the bottom of the page and check that users/projects are added

#### Types of changes

#### Checklist:


tansaku:
thanks @mattwr18 - so is this ready for standalone review?

mattwr18:
Yes, please @tansaku... please keep in mind that the reason we added the `params = "&infinite=true"` is because we are using `index.js.erb` in the users and projects views to handle both the `infiniteScroll` and the search functions. 
Not to say this is the best way of doing it, and if there is a cleaner way, I can refactor it. :smile:

tansaku:
no worries @mattwr18 - really great work splitting these out - will review ASAP - note you can always explicitly request a review from me on a github PR in order to make it completely clear that it's ready for my input:

https://help.github.com/articles/requesting-a-pull-request-review/

tansaku:
looks okay superficially - haven't been able to give detailed review yet

tansaku:
BTW I'm remembering how we used to have infinite scroll on users and it ground the site to a halt as the entire user list was loaded every time - hopefully the infinite scroll you've set up will be free of similar performance issues ...

mattwr18:
No pressure :sweat: I think it should be ok unless someone is scrolling like crazy :stuck_out_tongue: 
I'm getting excited to see it all in action!

tansaku:
me too ;-) although there are a lot of moving parts here so it might be easier to get them into production one at a time - which may hold off seeing them all working together :-/

I think the problem with the previous pagination was it relied on loading all existing users prior to displaying anything - your approach here looks much more performant.

mattwr18:
No problem, I'll be around when they all get merged in :P
The infiniteScroll is so fast sometimes I don't even see the `Please wait...` text before it is replaced by the users/projects/event instances

mattwr18:
hey @tansaku I merged develop into this branch, and adapted the tests to account for the search projects functionality. Can you have a look when you get a chance?

tansaku:
@mattwr18 will do - might need to get settled in the UK before I can really review properly :-) Sorry for the delay ...

mattwr18:
No worries, I'm not going anywhere :stuck_out_tongue_closed_eyes:

tansaku:
hi @mattwr18 - sorry for the delay on this - just starting to review.

All working for me locally - although I notice the explicit scroll links are still present, and also that the project width is reduced - i.e. it's now the width of the selector box, whereas on production it goes across as far as the search button:

in PR


vs

in production

this latter ^^^ from the current live site

tansaku:
thanks @mattwr18 - we can leave the page width if it will be changed in a subsequent PR.

Could we have the explicit pagination links only shown when javascript is off maybe?   I'd love to reduce the amount of stuff on the page, and it feels like an odd situation to have both present.  What happens with pagination if I have grabbed some extra items from the infinite scroll?  Should it page me forward a fixed amount, or the size of the set of projects showing from the infinite scroll?

mattwr18:
@tansaku I have fixed the width issue in the last commit, and I also removed the top pagination. The idea is that when a user scrolls down the bottom pagination is replaced and they will see a "Please wait..." text until the projects get populated, this happens until there is no "next" in the pagination. The idea is that the person will never see the pagination, or see for only a fraction of a second, but as soon as the window reaches a certain pixels to the bottom of the window it is replaced.

mattwr18:
@tansaku I followed this site to add this implementation: https://www.sitepoint.com/infinite-scrolling-rails-basics/. 
It failed the first build because of a code smell, which is when I realized someone else had added the same implementation for event instances and there was duplicate JS code, so I extracted that code into a function and called the function in the three places needed, /projects /users /hangouts

features/step_definitions/basic_steps.rb
```diff
@@ -420,12 +420,23 @@ def assert_link_exists path, text
...
     else
       expect(page).to have_content(project_title)
-    end  
+    end
+  end
+end
+
+Then(/^I should see "([^"]*)" within "([^"]*)":$/) do |project_title, project_list_area, table|
+  table.rows.each do |row|
+    step %Q{I should see "#{row[0]}" within "#{project_list_area}"}
```
roschaefer: Don't call step definitions from other step definitions. To DRY this code, you can simply define a method within this file, e.g. right above your step definitions that use it.

features/step_definitions/basic_steps.rb
```diff
@@ -420,12 +420,23 @@ def assert_link_exists path, text
...
 
+
+When(/^I scroll to the bottom of the page$/) do
+  page.execute_script "window.scrollBy(0,10000)"
+end
+
 Then(/^I should( not)? see "([^"]*)" within "([^"]*)"$/) do |negated, project_title, project_list_area|
   within("##{project_list_area}") do
     if negated
       expect(page).to_not have_content(project_title)
```
roschaefer: Never assert that sth. is not visible. This will always be true, e.g. if the page is not loaded. If you really have to, assert that sth. must be visible before checking that sth. is not visible. Right now, I don't know if `within('#id')` fulfills that.

features/step_definitions/basic_steps.rb
roschaefer: You don't make use of `project_title` here. Either say:
```ruby
 table.hashes.each do |row|
  title = row[project_title]
  # do sth. with it
end
```
Or even better hard code the `title` in the step definition. It is OK to have many step definitions.
```
Then I see the following projects in the list:
```

```ruby
 table.hashes.each do |row|
  title = row['Project Title']
  # do sth. with it
end
```

PR #2547, opened 2018-08-14
add other karma breakdown elements to karma table

### LINKED ISSUE #1295: add other karma breakdown elements to karma table
for consistency

vikasvr:
fixes #1295

Bit of renaming, memoization, few improvements which i think we could make here.
I wasn't clear with info i got from the above ticket, but i did this much with the understanding i had.

No tests added yet because i wanted to be sure that i haven't done something unnecessary.

@tansaku  @mattwr18  need your reviews :)

### DIFF
--- a/app/services/karma_calculator.rb
+++ b/app/services/karma_calculator.rb
@@ -6,16 +6,16 @@ class KarmaCalculator
     @user = user
   end
 
-  def perform
-    if user.karma
-      user.karma.total = 0
-    else
-      user.karma = Karma.find_or_create_by(user_id: user.id, total: 0)
-    end
-
-    return if user.created_at.blank?
-    user.karma.total = sum_elements
-    # better to have time in pairing sessions, code contributed (related to quality), issues, ...
+  def calculate
+    {
+      hangouts_attended_with_more_than_one_participant: number_hangouts_started_with_more_than_one_participant,
+      membership_length: membership_length,
+      profile_completeness: profile_completeness,
+      number_github_contributions: number_github_contributions,
+      activity: activity,
+      event_participation: event_participation,
+      total: sum_elements
+    }
   end
 
   private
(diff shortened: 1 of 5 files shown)


tansaku:
@vikasvr - this is looking good I think - been a while since I looked at the karma code so i could be missing things - but refactorings don't always need new tests if they are covered effectively by existing tests.

might be good to review the code coverage here to see if any needs tests

mattwr18:
@vikasvr I have a couple questions, I was wondering what purpose do the `{ }` in the calculate method serve?
Also, maybe I am missing something, but doesn't the `memoization` work to where if the value of the variable is `nil` then it will go the the right side of the `||=`, and then once it has a value it will not change it unless you explicitly assign it a different value? 
Maybe I am missing something obvious, but those were the two things I noticed that raised questions for me.
Thanks for getting this in the by way :smile:

vikasvr:
@mattwr18 
default value for any variable with @ sign is nil.
Thanx to Ruby :)

```
[1] pry(main)> @any_variable
=> nil
[2] pry(main)> @literally_any_instance_variable
=> nil
```
Before starting this i saw some code in karma calculator class and some in schedule task, which i thought something to clean up. So now the karma calculator just calculates the karma of user and returns a hash. And this hash acts exactly like the article_params, event_params which we use to update article and event model.

If we take user as example
```
user = User.last
user.first_name => nil
user.last_name => nil
user.attributes = {first_name: "Tony", last_name: "Stark"}
user.save
user.reload
user.first_name => "Tony"
user.last_name => "Stark"
```
And we can remove user.attributes and user.save by just calling user.update(first_name: "Tony", last_name: "Stark").

vikasvr:
Exactly

vikasvr:
It will return the recorded value as there no need to record the same thing again !

mattwr18:
I thought the idea is that there is a reason to calculate this value every day as part of the rake task? Like we don't want to calculate someone's activity once and then use that value as then there karma would never go up... It's not a static value, but something we expect to change quite often as a person contributes more and is around for longer, no?

mattwr18:
Also, where are these instance variables being used?

vikasvr:
It Will calculate it every day, the memoization is just for that particular cycle of the calculation of user karma. For example,
```
user = User.last
karma = KarmaCalculator.new(user)
for exactly this moment any instance variable say @membership_length is nil
now, when you call  karma.membership_length => it's value is calculated(say 5)
now, when karma. sum_elements is called,  this method ask for same membership_length value
which is now memoized so won't be calculated again.
But remember this memoization is just for this particular instance of the class.
For very next user , we will have different instance of the same class
Next day for same user , we will have different instance of the class
```

mattwr18:
Great explanation @vikasvr, thank you

tansaku:
great work @vikasvr - thanks for this - am going to merge and deploy

PR #2564, opened 2018-08-17
Add basic docs for setting up av dashboard token endpoint and giving access

mtc2013:
what it says


mtc2013:
@tansaku please review

PR #2579, opened 2018-08-22
Join button for Mob events visible to PremiumMob members and up

mattwr18:
Get the test passing which allow for PremiumMob members and up to see a JOIN THIS LIVE EVENT NOW button from the events show page for Mob events


app/models/user.rb
```diff
@@ -85,6 +85,15 @@ def current_subscription
     current_subscriptions.first
   end
 
+  def validate_user_subscription_plan(event)
+    if event.category == 'Mob'
+      subscription = current_subscription
+      subscription.plan.id > 2 unless subscription.nil?
```
tansaku: I get this although we are relying on the plan id sequence - fine for now, but might need something more robust in future ...

app/views/events/show.html.erb
```diff
@@ -112,14 +112,16 @@
   <% end %>
 
   <div class="col-xs-12 col-sm-6">
-    <% if @recent_hangout.try :live? %>
+    <% if @recent_hangout.try :live?  %>
+      <%  if current_user.nil? || current_user.validate_user_subscription_plan(@event) %>
```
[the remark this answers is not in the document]
mattwr18: That is what I started of with to get it passing, but then there were other tests failing. I like your method name and that it is a predicate method; although, can attend could be interpreted different than has attendance rights. 
I'll refactor this though and see if I can understand why it causes the other tests to pass, I guess the first place to start would be to add a visitor of the site to the acceptance test cause the whole point is that members with PremiumMob and up have attendance rights and below don't.

app/models/user.rb
```diff
@@ -85,6 +85,15 @@ def current_subscription
...
+    else
+      true
```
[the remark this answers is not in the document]
mattwr18: Haha, that is a much cleaner and more Ruby way of writing the same thing! HA

app/models/user.rb
mattwr18: I agree! It seems like a flaky way of doing things because it relies on my knowledge of the plan id and their corresponding levels and is not easily understood by others who haven't used pry to look at the inner workings. I'll have a think to see if I can come up with a better way.

mattwr18:
@tansaku got all the tests passing by adding a different button for people that can join Mobs... doesn't seem very elegant, but at least it's an option we could discuss.

tansaku:
great work @mattwr18 - so we could smooth the logic like so:

```erb
    <%  if @recent_hangout.try :live?  %>
      <% if @event.category == 'Mob' %>
        <% if current_user and current_user.allowed_to_attend?(@event) %>
          <% @message =  "IT'S TIME TO MOB NOW" %>
          <% @display = true %>
         <% else %>
          <% @display = false %>      
         <% end %>
       <% else %>
         <% @message =  "JOIN THIS LIVE EVENT NOW" %>
         <% @display = true %>
       <% end>
      <% if @display %>
      <div class="pull-right">
        <%= link_to @message, @recent_hangout.hangout_url, class: 'btn btn-success' %>
      </div>
     <% end %>
    <% elsif @event.try :within_current_event_duration? %>
      <div class="event-notice">
        It's time! Please start the event.
      </div>
<% end %>
```
but what we are starting to see here is too much business logic in the view, so that's an indication that we should start pulling that logic into the controller, and then probably elsewhere and then into a permissions framework like cancan, but one step at a time :-)

If it was me, I'd check that the above passes the tests and then refactor by pulling the variables @display and @message into the controller and thinking about renaming them ...

mattwr18:
hey @tansaku, I had to tweak it to get the tests passing, but I think they are all passing now... a few of us met cause it seems like a really good thing to pair on and get other people involved on, but we ran into issues with docker not working as expected on @aonomike and @tbaraza machine :cry: 
I reran docker on my end, no problems... not sure what's up, but it seems to have issues when people pull in a lot of changes :worried: 
We talked about pairing on refactoring this on Monday as soon as I get back from my mini trip. :smiley:

tansaku:
@mattwr18 is this ready for merge if review is good?

mattwr18:
We have created tests that check the functionality of the new feature @tansaku, and if it is sufficiently refactored, then it could be merged in; although, maybe we could have a better look cause I think there are things in the PR that aren't related to the feature.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180819232321) do
+ActiveRecord::Schema.define(version: 20180813125658) do
```
tansaku: I was trying to remove db/schema.rb from the repo ... wonder if we need to sync to develop?

app/views/events/show.html.erb
```diff
@@ -112,18 +112,20 @@
   <% end %>
 
   <div class="col-xs-12 col-sm-6">
-    <% if @recent_hangout.try :live? %>
+    <%  if @recent_hangout.try :live?  %>
+      <% show_live_events %>
```
tansaku: rather than setting an instance variable here, might it be simpler to have the show_live_events return a boolean?

tansaku: ah, I see you are setting the message as well ... hmmm

app/helpers/event_helper.rb
```diff
@@ -47,4 +47,21 @@ def google_calendar_link(event)
...
+      @display = false
+      if current_user and current_user.allowed_to_attend?(@event)
+        @message =  "IT'S TIME TO MOB NOW"
+        @display = true
+      end
+    else
+      @message =  "JOIN THIS LIVE EVENT NOW"
+      @display = true
+    end
+  end
```
tansaku: could this be:

```rb
  def show_live_events?
    return true unless @event.category == 'Mob'
    current_user and current_user.allowed_to_attend?(@event)
  end
```

app/views/events/show.html.erb
tansaku: if we remove that we can make the helper method much simpler ...

app/helpers/event_helper.rb
```diff
@@ -47,4 +47,21 @@ def google_calendar_link(event)
...
+
+  def event_frequency
```
tansaku: might we name this `repeats?` and perhaps move the code into the Event model?

mattwr18:
Right, @tansaku I think this is ready to get merged in? I stole all your good ideas, tweaked them to get the tests passing, and ran the build three times to get passed intermittent failures. 

Thank you to @StlMaris123 @faithngetich @aonomike as well as the `English Hack Night` and @RubyFloripa crew who helped get this functionality working! 

There are already some good suggestions on further features/tests that can be written in the features/events/private_events.feature file as comments. If anyone is keen to pair, let's do it!!

tansaku:
this is looking good @mattwr18 

I think there's one slight problem is that we're relying on a magic number corresponding to a database id in the app code - I pulled it out into a constant, but I think we still have a slight issue in that PREMIUM_MOB id on prod is 2 - but not in the tests.

maybe the quick fix is to ensure that the user had a plan that costs £25 or more a month - that's also a bit of a hack but a slightly more resilient one ... maybe checking that that user has a plan that is at least as expensive as premium mob would be be better.

In the long run we might need a whole role permissions framework, but I'd hate to have this feature be waiting for that ...

PR #2588, opened 2018-08-24
Update null user to have an id

### LINKED ISSUE #2583: Create an AnonymousUser type to simplify handling hidden users
Create an AnonymousUser or similar that would simplify user linking when user is anonymous, i.e. instead of this logic in the view:

```erb
          <% if project.user.id %>
            <%= link_to project.user.display_name, user_path(project.user) %>
          <% else %>
            anonymous
          <% end %>
        </div>
```

we could have an AnonymousUser entity have a display_name and an associated user page that explained what an anonymous user was and why we have them

NickSchimek:
#### Description
Updated the null user class to have an id.

This allows removal of the conditional in the view, because the
<link_to> method can now render since the returned NullUser object has
an id.

We should create an anonymous/ghost user in production with an id of -1.
(See https://github.com/AgileVentures/WebsiteOne/pull/2588#discussion_r223558623 for why the id of -1 is being used rather than a positive id # that's currently not being used in production)

Fixes #2583

#### Motivation and Context

This addition allows removal of code in the view.

tests already in place.

### DIFF
--- a/app/models/null_user.rb
+++ b/app/models/null_user.rb
@@ -5,7 +5,7 @@ class NullUser < User
   end
 
   def initialize(name)
-    super({ first_name: name, created_at: Time.now })
+    super({ id: -1, first_name: name, created_at: Time.now })
   end
 
   def presenter
--- a/app/views/projects/_listing.html.erb
+++ b/app/views/projects/_listing.html.erb
@@ -35,19 +35,15 @@
               <li data-toggle="tooltip" data-placement="top" title="<%= commit_count %> <%= 'commit'.pluralize(commit_count) %>">
                 <i class="fa fa-github-alt"></i> <%= commit_count %>
               </li>
+              <% end %>
             </ul>
-            <% end %>
           </div>
           <div style="display: inline-block; font-size: 1.0em; font-style: italic;">
             <div>Status: <strong style="font-style: normal;"><%= project.status.upcase %></strong></div>
             <div>Last update in Github: <%= project.last_github_update %></div>
             <div>
             <%= "Created #{time_ago_in_words(project.created_at)} ago" %> by
-            <% if project.user.id %>
             <%= link_to project.user.display_name, user_path(project.user) %>
-            <% else %>
-              anonymous
-            <% end %>
             </div>
             <div>
             <div class="row">
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -144,6 +144,9 @@ if StaticPage.count == 0
   Rake::Task['db:import_pages'].invoke
 end
 
+# Create ghost user
+Rake::Task['user:create_anonymous'].invoke
+
 klasses.each_with_index do |klass, i|
   puts "#{klass.name}.count " + old_counts[i].to_s.bold.red + ' -> ' + klass.count.to_s.bold.green
 end
--- a/features/projects/deactive_owners.feature
+++ b/features/projects/deactive_owners.feature
@@ -9,3 +9,11 @@ Feature: List projects with deactivated users
     When I visit "/projects"
     Then I should see "Home run"
     And I should see "by Anonymous"
+
+  Scenario: You can view the default anonymous ("Ghost") user page
+    Given the anonymous user exists
+    And "Billy Bob" creates the project "Home run"
+    When "Billy Bob" deactivates his account
+    And I visit "/projects"
+    And I click "Anonymous"
+    Then I should be on the anonymous profile page
--- a/features/projects/deactive_owners.feature
+++ b/features/projects/deactive_owners.feature
@@ -8,4 +8,4 @@ Feature: List projects with deactivated users
     And "Billy Bob" deactivates his account
     When I visit "/projects"
     Then I should see "Home run"
-    And I should see "by anonymous"
+    And I should see "by Anonymous"
--- a/features/step_definitions/projects_steps.rb
+++ b/features/step_definitions/projects_steps.rb
@@ -213,3 +213,8 @@ Given(/^"([^"]*)" deactivates his account$/) do |name|
   user = User.find_by first_name: first_name, last_name: last_name
   user.delete
 end
+
+Given(/^the anonymous user exists$/) do
+  attributes = { id: -1, first_name: 'Anonymous', last_name: '', email: 'anonymous@example.org' }
+  FactoryBot.create(:user, attributes)
+end
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -319,6 +319,10 @@ Given(/^I should be on the "([^"]*)" page for "(.*?)"$/) do |page, user|
   expect(current_path).to eq path_to(page, this_user)
 end
 
+Given(/^I should be on the anonymous profile page$/) do
+  expect(current_path).to eq('/users/-1')
+end
+
 Given(/^I (?:am on|go to|should be on) my "([^"]*)" page$/) do |page|
   page.downcase!
   if page == 'profile'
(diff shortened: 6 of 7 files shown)


NickSchimek:
yes, we'll need to create a user with an id of -1 in the db.

tansaku:
@NickSchimek makes sense - should we then be adding that in seeds.rb and possibly even creating a rake task to do that as well as maybe tests for that error and its resolution?

tansaku:
@NickSchimek did you think you'd have time to make the suggested changes before we pull this in? :-)

NickSchimek:
@tansaku Yes I do have time to work on this, and was waiting for you to get back from vacation as I had a couple questions for you that I was going to ask in a scrum later this week.

tansaku:
@NickSchimek did you have some questions about this one?

NickSchimek:
It's all done, but I want to check with you which email address I should put for the anonymous user. 
Do you want it to be an `@agileventures.org` address or a fake `@example.org` address?

tansaku:
@NickSchimek make it anonymous@agileventures.org and I'll forward it to myself

app/models/null_user.rb
```diff
@@ -5,7 +5,7 @@ def persisted?
   end
 
   def initialize(name)
-    super({ first_name: name, created_at: Time.now })
+    super({ id: -1, first_name: name, created_at: Time.now })
```
champagnepappi: @NickSchimek  just wanted to know. Why did you use a negative id instead of a positive one

NickSchimek: To prevent database errors.

Here's a little scenario: Let's say we have a fresh database. Never been used before. We want our ghost user to have an `id` of `1`. 
`Create command:`
```
[2] pry(main)> User.create id: 1, first_name: "ghost", email: "ghost@example.org", password: "password"
```
Database transactions are successful.
```
   (0.2ms)  BEGIN
  User Exists (0.6ms)  SELECT  1 AS one FROM "users" WHERE ("users"."id" != 1) AND "users"."slug" = $1 LIMIT $2  [["slug", "ghost"], ["LIMIT", 1]]
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = $1 AND "users"."deleted_at" IS NULL LIMIT $2  [["email", "ghost@example.org"], ["LIMIT", 1]]
  Authentication Exists (9.2ms)  SELECT  1 AS one FROM "authentications" WHERE "authentications"."user_id" = $1 LIMIT $2  [["user_id", 1], ["LIMIT", 1]]
  Authentication Exists (0.4ms)  SELECT  1 AS one FROM "authentications" WHERE "authentications"."user_id" = $1 LIMIT $2  [["user_id", 1], ["LIMIT", 1]]
  SQL (0.8ms)  INSERT INTO "users" ("id", "email", "encrypted_password", "created_at", "updated_at", "first_name", "slug") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"  [["id", 1], ["email", "ghost@example.org"], ["encrypted_password", "$2a$10$VD6der/ks6Y8WSsCF.ToT.cS44HL3NF9nn4foGitZS8qi56UtY1eG"], ["created_at", "2018-10-09 04:22:23.536777"], ["updated_at", "2018-10-09 04:22:23.536777"], ["first_name", "ghost"], ["slug", "ghost"]]
  ActsAsTaggableOn::Tagging Load (19.7ms)  SELECT "taggings".* FROM "taggings" WHERE "taggings"."taggable_id" = $1 AND "taggings"."taggable_type" = $2  [["taggable_id", 1], ["taggable_type", "User"]]
   (0.6ms)  COMMIT
  ActsAsTaggableOn::Tag Load (0.4ms)  SELECT "tags".* FROM "tags" INNER JOIN "taggings" ON "tags"."id" = "taggings"."tag_id" WHERE "taggings"."taggable_id" = $1 AND "taggings"."taggable_type" = $2 AND (taggings.context = 'skills' AND taggings.tagger_id IS NULL)  [["taggable_id", 1], ["taggable_type", "User"]]
  ActsAsTaggableOn::Tag Load (0.3ms)  SELECT "tags".* FROM "tags" INNER JOIN "taggings" ON "tags"."id" = "taggings"."tag_id" WHERE "taggings"."taggable_id" = $1 AND "taggings"."taggable_type" = $2 AND (taggings.context = 'titles' AND taggings.tagger_id IS NULL)  [["taggable_id", 1], ["taggable_type", "User"]]
```
Here is the return value from creating our ghost user:
```
=> #<User id: 1, email: "ghost@example.org", created_at: "2018-10-09 04:22:23", updated_at: "2018-10-09 04:22:23", first_name: "ghost", last_name: nil, display_email: nil, youtube_id: nil, slug: "ghost", display_profile: true, latitude: nil, longitude: nil, country_name: nil, city: nil, region: nil, youtube_user_name: nil, github_profile_url: nil, display_hire_me: nil, bio: nil, receive_mailings: false, country_code: nil, timezone_offset: nil, status_count: 0, deleted_at: nil, event_participation_count: 0, can_see_dashboard: false, skill_list: nil, title_list: nil>
```

Alright! Now that is out of the way let's have real users start signing up online.
Here's the first User: Notice I did not specify an id number, because the database increments these automatically.
```
[3] pry(main)> User.create first_name: "real_person", email: "real_person@example.org", password: "password"
```
Database transactions: Notice the Rollback and error message.
```
   (0.1ms)  BEGIN
  User Exists (0.5ms)  SELECT  1 AS one FROM "users" WHERE ("users"."id" IS NOT NULL) AND "users"."slug" = $1 LIMIT $2  [["slug", "real_person"], ["LIMIT", 1]]
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = $1 AND "users"."deleted_at" IS NULL LIMIT $2  [["email", "real_person@example.org"], ["LIMIT", 1]]
  SQL (6.5ms)  INSERT INTO "users" ("email", "encrypted_password", "created_at", "updated_at", "first_name", "slug") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"  [["email", "real_person@example.org"], ["encrypted_password", "$2a$10$yDOTvzvyxWpcG4x668EXneL5Loz9phWlDU.tnQzDrqGRlrUs.Qd06"], ["created_at", "2018-10-09 04:22:46.253559"], ["updated_at", "2018-10-09 04:22:46.253559"], ["first_name", "real_person"], ["slug", "real_person"]]
   (0.2ms)  ROLLBACK
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "users_pkey"
DETAIL:  Key (id)=(1) already exists.
: INSERT INTO "users" ("email", "encrypted_password", "created_at", "updated_at", "first_name", "slug") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"
from /Users/miafro/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.1.6/lib/active_record/connection_adapters/postgresql_adapter.rb:616:in `async_exec'
```

The database had a duplicate key value error, hence it tried to create the first signup using id of 1, but that was already taken because we manually added our ghost user to have an id of 1. 

So I can't use a positive integer except for `0`, because even if I use a large number say 1,000,000 eventually the database will get to the row number and error out.

so `-1` is a good value to use (`0` could be used too) unless negative numbers are being used for the database rows which they shouldn't unless there is a very good reason.

Bottom line is: Do not specify an id when manually creating database rows in production, but we needed to in order to roll out the ghost user without bugs so `-1` is our only option.

champagnepappi: @NickSchimek thank you for explanation. Makes more sense now

PR #2592, opened 2018-08-27
Embed video player, write test

### LINKED ISSUE #2584: Make event youtube links include an embedded easy to play video
The current list of youtube videos on the individual event pages look like this:

We have a similar pattern showing the videos associated with a particular individual, e.g. matt's personal page: https://www.agileventures.org/users/matthew-rider

and something similar for projects:

maybe it would make sense to copy over one of those templates such that we had the quick video play on the event page to make it more likely that folks might click on the embedded play on the event page itself ...

mattwr18:
Fixes #2584


tansaku:
oh cool - particularly cool if this does exactly what I think it does - any chance you can post a screen shot?

mattwr18:
@tansaku as I mentioned to you in one of the meetings today, I will need to create some EventInstances, but I can do that and send the screenshot

mattwr18:
screenshot at 2018-09-07 10 28 06
Maybe the video is too big?

tansaku:
thanks @mattwr18 maybe - think that's the same size as in projects, and profile pages - they could probably all do with being reduced a little to give folks a chance to see the list underneath ... but maybe we should get this deployed first.

Higher priority might be the bunched up text in the file list showing in your screenshot ...

mattwr18:
Actually, in the projects it is a much better size, but that is related to the container it's inside of I guess... 
https://www.agileventures.org/projects/websiteone#video

PR #2637, opened 2018-09-13
[WIP] Update relative paths to absolute for README pitches

### LINKED ISSUE #2636: Relative links in project's GitHub README.md are broken
## Expected Behavior
If there are relative links in a project's GitHub README file that point to other documentation within the same repo, those links should still point to the right place in the copy of the README that gets rendered on the AV project home page.

## Current Behavior
Links instead point to the same relative URLs on agileventures.org, so they are broken links. See for example:

* https://www.agileventures.org/projects/wiki-ed-dashboard - links in the Documentation section are broken
* https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/README.md same content with working link

## Context
New AV contributors are confused by the broken links.

mattwr18:
So that clicking on them from AV site leads to project's GitHub

fixes #2636

### DIFF
--- a/app/jobs/github_readme_files_job.rb
+++ b/app/jobs/github_readme_files_job.rb
@@ -28,7 +28,18 @@ module GithubReadmeFilesJob
 
   def project_readme(project)
     begin
-      project.update(pitch: content(project.github_repo, 'README.md'))
+      projects_readme = content(project.github_repo, 'README.md')
+      @doc = Nokogiri::HTML(projects_readme)
+      relative_paths = @doc.xpath('//a/@href').reject { |href| href.value.match?(/http/) || href.value.match?(/^\#/) }.map(&:value)
+      @absolute_paths = relative_paths.map { |relative_path| "#{project.github_url}/blob/master/#{relative_path}" }
+      i = 0
+      @doc.css("a").each do |link|
+        unless (link['href'].match?(/http/) || link['href'].match?(/^\#/))
+          link['href'] = @absolute_paths[i]
+          i += 1
+        end
+      end
+      project.update(pitch: @doc.to_html)
     rescue StandardError => e
       log_error(e, project)
     end


app/jobs/github_readme_files_job.rb
```diff
@@ -28,7 +28,18 @@ def run(projects)
...
+      relative_paths = @doc.xpath('//a/@href').reject { |href| href.value.match?(/http/) || href.value.match?(/^\#/) }.map(&:value)
+      @absolute_paths = relative_paths.map { |relative_path| "#{project.github_url}/blob/master/#{relative_path}" }
+      i = 0
+      @doc.css("a").each do |link|
+        unless (link['href'].match?(/http/) || link['href'].match?(/^\#/))
+          link['href'] = @absolute_paths[i]
+          i += 1
+        end
+      end
+      project.update(pitch: @doc.to_html)
```
tansaku: this below will probably not work as is, but we maybe could try this approach

```rb
RELATIVE_PATHS_REGEXP = /href="(((http|https|ftp)://)|\#)(^")*"/
RELATIVE_LINK_GROUP_ID = "$3"

absolute_path_prefix = "#{project.github_url}/blob/master/"

projects_readme.gsub!(RELATIVE_PATHS_REGEXP, "href=\"absolute_path_prefix#{RELATIVE_LINK_GROUP_ID}\"")
```

tansaku:
scratch of a possible unit test:

```rb
  describe '.replace_relative_links_with_absolute' do
    let(:project){ double Project }
    let(:project_readme_content) {  '<a href="hello">hello</a> <a href="http://example.com/hello"</a>hello<a>'}
    let(:converted_readme_content) {  '<a href="http://github.com/AV/p/hello">hello</a> <a href="http://example.com/hello"</a>hello<a>'}
    it 'converts all relative urls in text to absolute with full host prefix' do
      converted_text = subject.replace_relative_links_with_absolute(project_readme_content, project)
      expect(converted_text).to eq converted_readme_content
    end
  end
```

mattwr18:
Hey @tansaku, I added some unit tests to check the functionality of that method. 
I needed to take a different route because when we parse the html with Nokogiri the output is much longer.
I think the method I used is equally testing what we want, but cleaner.

PR #2643, opened 2018-09-18
Automate vcr renewal

mattwr18:
Refresh vcr cassettes, tests
Add automatic renewal of vcr cassettes


mattwr18:
@tansaku it is failing the build because the total-coverage, but seeing as the twitter feature is not an active feature, I removed it... Is this what you were talking about wanting? It seems like it will update our vcr cassettes once a month. More or less frequently, if you want. 

The thing about this is that their is still a manual component to this, which is updating the GitHub devops tests, or maybe we can not renew them?

PR #2693, opened 2018-10-15
sorting github pages snarl up

### LINKED ISSUE #2630: premium static page snarl up
see https://github.com/AgileVentures/AgileVentures/issues/57 for details, but solution will require coordination between both repos ...

tansaku:
have started with adding cuke for github static pages import

we also need to work out the underscores vs hyphens issue ...

ultimately fixes #2630 
related to #AgileVentures/AgileVentures/issues/57


mattwr18:
Great @tansaku! you are fast, thanks for this... it will definitely help us to carry on working on this... I wonder if we shouldn't be checking in the VCR yet, maybe we could merge in the changes to filter out the sensitive data and then commit the cassette?

tansaku:
oh yes, sod, I assumed the github client secret and id would be being filtered already - that needs fixing - I wonder if all the other github related tasks are in the same situation ...

PR #2696, opened 2018-10-15
[WIP] 2676 run related tests when committing

### LINKED ISSUE #2676: run related tests when committing
would be great to have it commit hooks that ran tests related to changes made

mattwr18:
@tansaku let's talk about the pros and cons of this in a scrum?
Maybe we would actually want to just run a custom script based on files changed to speed it up
Also, doesn't seem to support cucumber tests out of the box. 
Surely, we would need to enable and disable the hooks based on our needs

Fixes #2676


mattwr18:
@tansaku is this still desired?

tansaku:
thanks for looking into this @mattwr18 - hopefully we can revisit when we have more folks working on the project :-)

PR #2711, opened 2018-10-16
refactor cucumber sponsors banners

### LINKED ISSUE #2705: refactor cucumber
Robert suggested:

> I would repeat my suggestions from this pull request please don't use cucumber as a mere test runner. Cucumber is documenting the behaviour of your app. I like Then I see the banners of all sponsors from above as a cucumber step. You can store the list of sponsors in your step definitions with an instance variable @sponsors = ... whatever

for the sponor cukes

dkuku:
Refactored the tests as suggested, I used a local variable to keep the data with the test

fixes #2705

#### Description

Used a hash for checking the img => link dpendency

#### Motivation and Context

Simplifies current feature file

By running this file locally, I also checked if there is an error it the values differ.
root@7a0f96985260:/WebsiteOne# bundle exec cucumber features/pages.feature
8 scenarios (8 passed)                                                                                              
52 steps (52 passed)
0m4.025s

#### Types of changes

#### Checklist:


features/step_definitions/pages_steps.rb
```diff
@@ -63,4 +63,18 @@ def static_page_path(page)
 Then(/^the "([^"]*)" page title should read "([^"]*)"$/) do |str, title|
 	visit path_to(str)
 	expect(page.title).to eq "#{title}"
-end
\ No newline at end of file
+end
+
+Then(/^I see the banners for all sponsors$/) do
+sponsors = {
```
diraulo: Fix indentation here

features/step_definitions/pages_steps.rb
```diff
@@ -63,4 +63,18 @@ def static_page_path(page)
...
+  "Standuply" => "https://standuply.com/?utm_source=links&utm_medium=agileventures&utm_campaign=partnership",
+  "Craft Academy" => "http://craftacademy.se/english",
+  "Mooqita" => "https://mooqita.org/",
+  "Amazon Smile" => "https://smile.amazon.co.uk/ch/1170963-0"
+}
+  sponsors.each do |key, value|
+    within("//a[@href='#{value}']/div") do
+      find(:xpath, "//img[@alt='#{key}']")
```
diraulo: Using `find` here I suppose the test will fail if it doesn't find the image on the page right?
It's not immediately clear while reading this that we expect to find an image using this test ... or perhaps it's just me. I'd probably have something along the line of

```ruby
expect(page).to have_selector("img[@alt='#{key}']")
```
What do you think about that?

features/step_definitions/pages_steps.rb
dkuku: I had it outside the method and pasted this and forgot about the indentation

features/step_definitions/pages_steps.rb
dkuku: Thanks @diraulo for the review, I also need to change the key-value variable for something which makes more sense.

tansaku:
thanks so much @dkuku ! :-)

PR #2743, opened 2018-10-19
Removed rack-timeout from development env, fixes 2686

### LINKED ISSUE #2686: Rack Timeout in development environment prevents use of debugging tools
Drop in a binding.pry and try to call it and you will not be able to use your debugging session because of rack timeout warnings.

Can we turn off rack timeout for development environment?

dkuku:
fixes #2686
Removed rack-timeout from development env - now it's possible to use pry and byebug
In my opinion we should not use it in development anyway. From rack-timeout documentation:

> Rack::Timeout is not a solution to the problem of long-running requests, it's a debug and remediation tool. App developers should track rack-timeout's data and address recurring instances of particular timeouts, for example by refactoring code so it runs faster or offsetting lengthy work to happen asynchronously.

Also added rdoc gem to Gemfile, Without this I could not install this app locally on linux manjaro due to missing `rdoc/markdown/to_html` file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -119,9 +119,10 @@ group :development, :test do
   gem 'rspec-html-matchers'
   gem 'rspec-rails'
   gem 'simplecov', '~> 0.16.1'
+  gem 'rdoc'
 end
 
-group :development, :staging, :production do
+group :staging, :production do
   gem 'rack-timeout'
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -119,7 +119,6 @@ group :development, :test do
   gem 'rspec-html-matchers'
   gem 'rspec-rails'
   gem 'simplecov', '~> 0.16.1'
-  gem 'rdoc'
 end
 
 group :staging, :production do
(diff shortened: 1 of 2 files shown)


mattwr18:
hey @dkuku this is amazing!! I had a quick search on the internet, but hadn't thought about looking in the Gemfile to see what group `rack-timeout` was in... good catch! did you see it's failing the build with this message?
```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

The dependencies in your gemfile changed

You have added to the Gemfile:
* rdoc
```

dkuku:
I need to learn to look at the ci output - any way to have it configured to message me when it's failing ?
So I don't have to wisit this page every time to check. Or maybe it is but I disabled the notifications?

mattwr18:
@dkuku It would be great to configure it to post the relevant failures in the comments section! I think there is a way, but haven't really looked that much into it. 
We do have a channel #websiteone-notify that tells you about all the activity, so you'll see when someone comments, merges, etc.
It's for all PRs though, not specific to yours

PR #2816, opened 2018-11-08
removed nearest_time_zone gem and all references

### LINKED ISSUE #1287: replacing nearest_time_zone
could improve memory performance - before doing full replace would be nice to see how removing it would impact production

dkuku:
#### Issue addressed
fixes #1287

#### Testing
removed all tests dependand on this functionality, all other all passing

####TODO
There is a migration file whick created a column in users table connected with this functionality, should I create another one for removeing the timezone_offset column ot this one can be reversed ? 
`db/migrate/20150208124239_add_timezone_offset_to_users.rb`


dkuku:
I migrated the database to remove the timezone_offset from User model and, the tests and these are still passing

tansaku:
wonderful work here @dkuku ! :-)

PR #2824, opened 2018-11-12
adds query caching for upcoming events

### LINKED ISSUE #2118: improve events page performance
currently takes several seconds to load the events index page - from previous analysis my sense is that the time is spent in the ruby code generating the complete list of upcoming events from the recurring events.

We could improve performance by storing horizon of event data in the db up to six weeks and recalculating on edit, not on view ...

tansaku:
#### Issue addressed

fixes #2118 

Not currently working:

#### Testing

hoping to test this manually locally, but error blocking that at the moment ...


mattwr18:
screenshot at 2018-11-13 09 34 31
It says it is reading the fragment, but the file it says it's reading from isn't being created locally like it's says in the edgeguides 
```
When your application receives its first request to this page, Rails will write a new cache entry with a unique key. A key looks something like this:
views/products/1-201505056193031061005000/bea67108094918eeba42cd4a6e786901
```

tansaku:
thanks for this @mattwr18 - I think it's not working because we're caching after the expensive operation is already peformed.  There's no improvement on production like data, and users/projects load fast for me, while events is still super slow - I think we need https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching to cache the actual method call ...

tansaku:
although the expiration date is a problem, but even with:

```rb
  def index
    @projects = Project.active
    @events = Rails.cache.fetch("events", expires_in: 12.hours) do
      Event.upcoming_events(specified_project)
    end
    respond_to do |format|
      format.html
      format.json
    end
  end
```

it doesn't work ... that's in combination with the fragment caching ...

tansaku:
tried this as well - didn't work ...

```rb
  def self.upcoming_events(project=nil)
    Rails.cache.fetch("events", expires_in: 12.hours) do
      events = Event.base_future_events(project).inject([]) do |memo, event|
        memo << event.next_occurrences
      end.flatten.sort_by { |e| e[:time] }
      Event.remove_past_events(events)
    end
  end
```

mattwr18:
Currently failing with these tests:
Failures:

  1) Event#upcoming_events does not show finished events
     Failure/Error: expect(Event.upcoming_events.count).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/models/event_spec.rb:495:in `block (3 levels) in <main>'

  2) Event#upcoming_events returns event 1 minute before ending
     Failure/Error: expect(Event.upcoming_events.count).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/models/event_spec.rb:500:in `block (3 levels) in <main>'

  3) Event#upcoming_events shows future events
     Failure/Error: expect(Event.upcoming_events.count).to eq(2)
     
       expected: 2
            got: 0
     
       (compared using ==)
     # ./spec/models/event_spec.rb:490:in `block (3 levels) in <main>'

Finished in 1 minute 0.6 seconds (files took 3.8 seconds to load)
656 examples, 3 failures

Failed examples:

rspec ./spec/models/event_spec.rb:493 # Event#upcoming_events does not show finished events
rspec ./spec/models/event_spec.rb:498 # Event#upcoming_events returns event 1 minute before ending
rspec ./spec/models/event_spec.rb:488 # Event#upcoming_events shows future events
```
Interestingly, locally these are passing intermittently and exclusively(?) when not run in isolation @tansaku

**Comment by tansaku** [comment]
failing with:

```
Failing Scenarios:
cucumber -p first_try features/events/create_events.feature:69 # Scenario: Create a new event in a non-UTC timezone
cucumber -p first_try features/hangouts/hangout.feature:131 # Scenario: Infinite scroll on hangouts scroll down until no more hangouts
cucumber -p first_try features/hire_me_modal.feature:12 # Scenario: Sending a message to user with 'Hire me' button
```

tansaku:
@mattwr18 passed now - cool - but I think we might have to add background jobs to handle the editing of the individual events:

https://guides.rubyonrails.org/active_job_basics.html

and then we'd need something like Sidekiq - some more details here:

https://devcenter.heroku.com/articles/background-jobs-queueing

worried about how involved it might be to set up ...

PR #2826, opened 2018-11-13
Cache 304 responses for events

### LINKED ISSUE #2118: improve events page performance

mattwr18:
This caches all the events and if there it receives a 304 response loads from cache
#### Issue addressed

fixes #2118 

first request
screenshot at 2018-11-13 20 37 23
caching info and subsequent request
screenshot at 2018-11-13 20 43 51

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source 'https://rubygems.org'
 ruby '2.5.1'
 
 gem 'rails', '~> 5.1'
-gem 'actionpack-action_caching'
 gem 'acts-as-taggable-on'
 gem 'acts_as_follower', git: 'https://github.com/AgileVentures/acts_as_follower.git'
 gem 'acts_as_tree'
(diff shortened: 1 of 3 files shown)


mattwr18:
@tansaku I created a separate branch for this to test it in isolation... could you please pull this down and test it with the production like data?

mattwr18:
For me in development mode and with much less data the thing that is taking the longest to load is the assets which are usually coming back with a 304 response meaning we could be caching them as well, haven't worked out exactly how to do that yet, and also in production that is not the biggest problem. 
screenshot at 2018-11-13 20 50 45
this above was one of the faster times... it was taking 7s earlier

mattwr18:
Have made the changes we discussed in the Scrum, changing `@events = Event.all` to `@events = Event.upcoming_events(specified_project)` and the tests pass, but there is no savings in time to return a response on events

mattwr18:
Or maybe it has, hard to tell without production like data.. I'll push it up

PR #2836, opened 2018-11-15
[WIP] admin rake tasks

### LINKED ISSUE #2814: rake tasks for premium admin actions
Each month the admin needs to review the Premium members and check that they are all in sync, i.e. the data on their plans etc. is all correct.

At the moment the Admin does this by using the rails console and some little notes on how to connect up accounts etc.  Ideally there would be an admin interface for this, but in the first instance a few rake tasks would allow us to get the ruby scripts in version control and could then be run on production from the command line a little more easily ...

tansaku:
Very early days for this one

#### Issue addressed

fixes #2814

### DIFF
--- a/lib/tasks/admin/premium.rake
+++ b/lib/tasks/admin/premium.rake
@@ -18,6 +18,16 @@ namespace :util do
     time = DateTime.parse(datetime)
     AddSubscriptionToUserForPlan.with(user, user, time, plan, payment_source)
   end 
+  desc 'give Premium mob subscription via sponsor'
+  task :paypal, [:email, :sponsor_email, :datetime, :identifier] => [:environment] do |task, args|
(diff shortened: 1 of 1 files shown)


mattwr18:
Can help you with this @tansaku, if you want. I might need some context... would it be quicker for you to write tests that I can get passing? or would that take you longer than just implementing this?

tansaku:
@mattwr18 I'd love to look at this together - maybe as part of next week?  we need sprint planning ...

mattwr18:
hey @tansaku I know you are busy, but just wanting to know if we can make some headway on this?

tansaku:
good question @mattwr18 - I'm going to try and document what I do now in the #mentors-devops channel on slack

PR #2855, opened 2018-11-18
Event doesnt default to a project when not selected

### LINKED ISSUE #2854: Events created with no project select should not default to any project
When a user creates a new event, and decides not to associate it to any project, no project should be associated to the given event, unless cs169 is among the list of projects

## Expected Behavior

When an event is created and no project is linked then no project should be linked to the event in the database
## Current Behavior

Currently, in  a case where cs169 is not one of the projects available, when an event is created and no project is selected, the project defaults to the first project in the list in the order of alphabet of the project names
## Possible Solution

Make the event not default to any project if the user does not select a project from the projects dropdown when creating an event

1. Ensure that cs169 is not among the projects you have available
2. Navigate to the create events page /events/new 
3. Fill in all fields except Project drop-down
4. Click save

## Context

When working on feature #1022 I noticed one reason projects events are defaulting to wrong projects on update was because it was being set to a wrong project on creation

aonomike:
Add test to check to check that when an event is created and no project is assigned by the user, a default project is not assigned
#### Issue addressed

fixes #2854


features/step_definitions/event_steps.rb
```diff
@@ -467,3 +467,11 @@ def stub_user_browser_to_specific_timezone
 And(/^I hit back$/) do
   page.go_back
 end
+
+Then(/^the event named "([^"]*)" is not associated with any project$/) do |event_name|
+   # Write code here that turns the phrase above into concrete actions
+   event_project_id = Event.find_by(name: event_name).project_id
+   Project.all.pluck('id').each do |project_id|
```
tansaku: no need to loop - just check the project id is nil

PR #2903, opened 2018-12-02
fix for pagination test

dkuku:
In the before action there are created 2 hangouts, then in last test we are creating additional 18 which is 20 in total.
The problem I had when the sorting order changed this test started to fail - on second page It was showing 9 instead of 10 because one had different title and step definition was not counting it as a hangout.
The Step definition checks if the name starts with "Hangout_" - this probably needs to be updated as well for something more generic.
 
I renamed the first two hangouts. Now the test checks if there are 20 hangouts, which were there before but two of them had different name, if we would create 19 hangouts instead 18 the test would fail.


dkuku:
@tansaku the test looks on my computer like this: 
```
    @javascript                                                                                                                                                                                                 
      Scenario: Infinite scroll on hangouts scroll down until no more hangouts                                                                                                                                    
      Given 16 hangouts exists                                                                                                                                                                                  
       When I visit "/hangouts"                                                                                                                                                                                  
       Then I should see 6 hangouts                                                                                                                                                                              
       And I scroll to bottom of page                                                                                                                                                                            
       Then I should see 12 hangouts                                                                                                                                                                             
       And I scroll to bottom of page                                                                                                                                                                            
       Then I should see 18 hangouts                                                                                                                                                                             
       And I scroll to bottom of page                                                                                                                                                                            
       And I should see "No more hangouts" 
```
on the CI it fails because there is 20 instead of 18, (we create 16 and there are 2 from the before block)
and git does not pick it up because I have 18 instead 20 in this branch. I probably messed it up somehow when I created the branch.

dkuku:
O god - I'm sitting on it 1 hour and I know whats the problem is - I posted this before I done the video wall.
Then in the video wall branch it was further updated to paginate by 6 not by 5. So this is not relevant anymore.

tansaku:
thanks @dkuku - I'll close this then

PR #2929, opened 2018-12-10
Append error message to array

### LINKED ISSUE #2857: Blocked on creating new projects
## Expected Behavior

Should be able to create a new project
screenshot at 2018-11-18 15 52 39

## Current Behavior

Getting a 500 Internal Server Error

NickSchimek:
## Description

The messages hash has a value listing all errors in an array. Updated
Image validator to append error messages to the array instead of over
writing the array with a String. This is being done because it causes
Rails ActiveRecord to error out when it does post validation checks,
because it's expecting an array, but instead receives a String.

Fixes #2857


tansaku:
great detective work @NickSchimek :-)

PR #2989, opened 2019-01-08
[WIP] Devise auth token

mattwr18:
#### Issue addressed

fixes #


mattwr18:
This is not working, and I'm not sure why @tansaku... users are able to sign up using devise still, but sending a post request to the http://localhost:3000/api/v1/auth endpoint which is supposed to create a user, as far as I understand it, is returning
```
{
    "success": false,
    "errors": [
        "undefined method `each' for #<Proc:0x00007fdbbc77a700>"
    ]
}
```
As far as I understand, there is no need to maintain Devise as the frontend will be replaced by React, but I am not sure what steps to take...

mattwr18:
there is also this that we need appartently https://github.com/mcordell/grape_devise_token_auth

PR #2994, opened 2019-01-13
[WIP] Add devise jwt

### LINKED ISSUE #2993: Add support for devise-jwt
Allows for authentication via api 
Will use for WebsiteOne React Frontend

mattwr18:
Add the gem and configure
#### Issue addressed

fixes #2993 

#### Testing

has not been tested yet


mattwr18:
I added a [WIP] tag @tansaku cause these changes cause the tests to break... I was hoping that we could get these changes merged in to develop, especially since we might need to turn off the avp box where the webisteone-develop lives, but I think it would require changing the registrations/sessions and I'm sure that we could support logging in through the current site and through an api/ our react frontend, but I'm not sure exactly how to do that nor am I sure whether it's worth spending the time to do that as the idea is to replace the frontend with react... any advice on a way forward would be appreciated.

mattwr18:
maybe this should be closed @tansaku? we can talk better in a future meeting maybe? 
not sure if the issue is clear...

tansaku:
thanks @mattwr18 - I'm not completely following it - but I see all the failing tests are request specs.  If the cucumber tests all pass I might be tempted to delete those request specs.  My intuition is that the controller specs are generally binding too tightly and not allowing us the flexibility we need.

Of course I might be wrong and the cucumber tests may be vacuous and the request specs are actually checking something important, but difficult to know for sure without a detailed investigation - question is whether spending that time on investigating is even worth it, when we could test by just deploying the branch to develop or perhaps even better staging and seeing if authentication works ...

PR #2997, opened 2019-01-14
2652 paypal bug workaround

### LINKED ISSUE #2652: premium mob sign up got 404 after returning from paypal
that's all I have at the moment - not a good user experience - needs review

mattwr18:
this is a workaround for what seems to be a bug from paypal, where a user is returned to the site with a get request instead of a post with like previously
#### Issue addressed

fixes #2652 

screenshot at 2019-01-14 18 27 48

#### Testing

we have added a test to uncover the bug and then implemented the fix


features/premium/subscribe_self_to_premium_mob.feature
```diff
@@ -47,3 +47,11 @@ Feature: Subscribe Self to Premium
...
     Then "sam-buyer@agileventures.org" should not receive a "Welcome to AgileVentures Premium Mob" email
     And I should see "redirected" in last_response
+
+  Scenario: PayPals redirect doesn't work
+    Given I have logged in
+    And I visit "subscriptions/new?plan=premiummob"
+    Then I should see a paypal form within the paypal_section
+    When Paypal updates our endpoint for premium mob via get
+    Then "sam-buyer@agileventures.org" should receive a "Welcome to AgileVentures Premium Mob" email
+    And I should see "Thanks, you're now an AgileVentures Premium Mob Member!" in last_response
```
faithngetich: Great work!! You might want to add a new line here

tansaku:
great work @mattwr18 @FedericoEsparza !

features/premium/subscribe_self_to_premium_mob.feature
mattwr18: great @faithngetich thanks :blush:

app/views/subscriptions/new.html.erb
```diff
@@ -50,7 +50,7 @@
         <input type="hidden" name="no_shipping" value="1">
 
         <!-- url to return to on success -->
-        <input type="hidden" name="return" value="<%= paypal_return_url %>">
+        <input type="hidden" name="return" value="<%= "#{paypal_return_url}?item_number=#{current_user.slug}&item_name=#{@plan.name}&auth=#{cookies['_WebsiteOne_session']}" %>">
```
tansaku: what would be awesome to keep the tests and this in sync, would be making this a helper method that we could also refer to in the test, that way we'd be sure the test was assembling the return_url the same way the form is

features/step_definitions/premium_steps.rb
```diff
@@ -84,7 +84,18 @@ def submit_card_details_for_button_with(text, email='random@morerandom.com', num
...
+                        'last_name' => 'buyer', 'residence_country' => 'GB', 'mc_currency' => 'GBP',
+                        'business' => 'sam-facilitator@agileventures.org', 'recurring' => '1',
+                        'payer_status' => 'verified', 'first_name' => 'test',
+                        'receiver_email' => 'sam-facilitator@agileventures.org',
+                        'payer_id' => '9EG5X4H5DJJW4', 'reattempt' => '1',
+                        'item_number' => 'not logged in', 'subscr_date' => '10:07:19 Dec 12, 2016 PST',
+                        'charset' => 'windows-1252', 'period1' => '7 D', 'mc_amount1' => '0.00',
+                        'period3' => '1 M', 'mc_amount3' => '10.00',
+                        'auth' => 'A31jSI5vY44zpPcQlAUk8WdibsJJT72rGx6ptiGPil6MG30OuCoFtHJ38.CJmmBQ.NNbZg.XEaWj298bVa5FZIw',
+                        'form_charset' => 'UTF-8'}
```
tansaku: big improvement - any chance we could get every item on it's own line?

PR #3000, opened 2019-01-16
Update api endpoints

mattwr18:
didn't create an issue for this
with the difficulties deploying to avp box, it brings up the fact that these changes could get merged in
@tansaku, what do you think?


tansaku:
I could have sworn we had already merged these in - at least I strongly recall seeing all these codeclimate issues and you putting in a fix for them in the past - I don't know why all this isn't merged in already - feels like we lost commits somewhere

Also, I think it would be great to have this connected to an issue, e.g. supporting the react front end, so we could connect this and future PRs to it in one place ...

mattwr18:
@tansaku closing this for now... the api endpoints have been updated since this `PR` and will change again in the future... some of the changes can be merged into develop and pushed along the pipeline to production, but others like the devise-jwt would need to be modified to support both json requests and requests coming in from the rails interface for them to be mergeable... 

maybe we can talk in a future meeting on what the transition to the React front end would even look like?

PR #3025, opened 2019-01-25
ensure new projects events are posted to slack

### LINKED ISSUE #2802: ensure that new members events are posted to slack #new_members channel
currently there is no way to get automated posts to the #new_members channel - currently we hard code to post to the channels of the project that an event is associated with ... could create a "New Members" project?  Or some sort of over-ride to have events associated with slack channels independently of events ...

mattwr18:
this is not an ideal way of dealing with this issue, in my opinion because it is increasing the number of hardcoded slack ids
this is a stop-gap measure until we can prioritize pulling in the values from the backend 
#### Issue addressed

fixes #2802


tansaku:
great work @mattwr18 - as you say not ideal, but a stop gap - this is a key area - funny, but I start to think about microservices again as a way to quickly get these things out - ironically given I merged the old microservice into our monolith - but either way our problem is resources - we just need people willing to spend time programming on the issue, or having funds to pay them ... we have a partial fix with @ElisaRmz 's initial work - I don't think it would take more than a good day's concentration to geta more generic fix ... probably want that before going crazy with microservices ...

PR #2975, opened 2019-01-02
Refactor Slack Notifications

ElisaRmz:
In this PR I want to refactor the notifications code before add the new notifications for slack asked in #2802

PR #3077, opened 2019-03-05
[WIP] switch to using project.slack_channels instead of hard coded

### LINKED ISSUE #2802: ensure that new members events are posted to slack #new_members channel

tansaku:
Use slack channels from db AND allow events to be associated with slack channels

#### Issue addressed

fixes #2802


tansaku:
@marianmosley thanks for pushing that up - great work ...

hmm, so to completely fix the issue for pinging the new members channel we'd also need to set up a many to many  relation between events and slack channels, and also ultimately provide an interface to allow event slack channel relations to be set ... or we hack things with a new_members project ...?

mattwr18:
@tansaku, we were having a look at this in the Agile Dev Mob today, and we had a doubt that maybe you can address? 
Could you explain what the purpose of have a join table for `project.slack_channels`?
it was merged in as part of this `PR` https://github.com/AgileVentures/WebsiteOne/pull/2975

Ryder brought up, and I tend to agree with him, that a project should only have one `slack_channel`, at the moment there is an attribute `slack_channel_name` that allows a user to add a slack_channel when they create a project. Should we not be just adding an attribute `slack_channel_code`, if needed, to ping the correct slack channel?

We understand that an event may have the use case of wanting several slack channels to be pinged, or may not be associated with a project, but be associated with a slack channel, and therefore could have an array of slack channel codes, which would justify having a join table.

Please let me know if there is something I am missing about the use case.

tansaku:
@mattwr18 many of our projects have multiple slack channels, e.g. websiteone has #websiteone and #websiteone-notify and #websiteone-deveops - @ElisaRmz brought that join table in.

The way we were developing this is that when an event started we would look for both the set of slack channels associated with the project the event is associated with, and any slack channels the event was associated with.

> Should we not be just adding an attribute slack_channel_code, if needed, to ping the correct slack channel?

I'm not sure what the different between `slack_channel_code` and `slack_channel_name` is.  To be honest we could probably have gotten away without a join table between slack channels and projects, but I see no particular harm in having one or doing the relation that way, since that's how @ElisaRmz started.  Right now I'm focusing on the event/slack_channel relation (which this PR introduces) in order to make it possible for events not associated with projects to ping a number of channels (new members meetings, retros, marketing etc.)

PR #3123, opened 2019-04-02
refactor slack channel notifications use projects slack channels join table

### LINKED ISSUE #3118: Refactor to use project.slack_channel_codes
## Expected Behavior

Hangout/YouTube notification services should use `slack_channels` join table to ping slack based on a project's slack channels
## Current Behavior

At the moment, the values are hard coded in making it more difficult to maintain 
## Possible Solution

It would be nice to have this automated in the future, but for now this issue is just to transition to move away from hard coded values in the code base

mattwr18:
#### Issue addressed

fixes #3118


tansaku:
1) should we also add code to do same for the youtube service?  YES
2) should we create a one-shot rake task to run through the ChannelList.rb file and populate the projects_slack_channels table? YES

lib/tasks/db.rake
```diff
@@ -5,4 +5,18 @@ namespace :db do
       project.source_repositories.create(url: project[:github_url])
     end
   end
+
+  desc "add slack channel codes to project slack channels join table"
+  task add_project_slack_channels: :environment do
+    include ChannelsList
+    Project.all.each do |project|
+      if CHANNELS[project.slug.to_sym].respond_to? :each
```
tansaku: in principle we could use `Array()` here to simplify this code, but we'll only use it once, so if it's working it's almost not worth changing ...

tansaku:
great work @mattwr18 - you said you tested this in Azure, so we could merge this and take it all the way to production perhaps?

mattwr18:
yes, let's get it merged...

PR #3136, opened 2019-04-12
Add event slack channels

### LINKED ISSUE #3135: refactor slack channel notifications use events slack channels join table
## Expected Behavior

we would like to be able to save an event's slack channel code in a join table to be able to ping slack channels associated with an event
## Current Behavior

only allows for an event to be associated with a slack channel through its project
## Possible Solution

create a join table with a `has_and_belongs_to_many` relationship between the event and slack channel models

mattwr18:
- Add associations
- Add join table migration
- Add models test
- refactor notification services tests

Co-authored-by: Gideon Kimutai <gideon.kimutai.kim@gmail.com>

#### Issue addressed

fixes #3135 

#### Testing

Not really happy with the youtubenotificationservice spec... 
the easiest thing to show proof of concept was to add it to an existing test entitled `sends the correct slack message to the correct channels`, but this does not help for others to understand what we are testing
also, I worry about having three expectations in the same it block... 
@tansaku please have a look and see what you think
we talked about a rake task to add event.slack_channels, but it's a bit different... I could come up with a hardcoded list of recurring events, which are not pinging the channels we expect?

### DIFF
--- a/app/models/event_instance.rb
+++ b/app/models/event_instance.rb
@@ -64,7 +64,7 @@ class EventInstance < ApplicationRecord
     yt_video_id && "https://youtu.be/#{yt_video_id}"
   end
 
-  def channels_for_event
+  def channels_for_event event_id
     return [] if event_id.nil?
     event = Event.find(event_id)
     event.slack_channel_codes
--- a/app/models/event_instance.rb
+++ b/app/models/event_instance.rb
@@ -64,9 +64,9 @@ class EventInstance < ApplicationRecord
     yt_video_id && "https://youtu.be/#{yt_video_id}"
   end
 
-  def channels_for_event event_id
-    return [] if event_id.nil?
-    event = Event.find(event_id)
+  def channels_for_event
+    return [] if self.event_id.nil?
+    event = Event.find(self.event_id)
     event.slack_channel_codes
   end
 
--- a/app/models/event_instance.rb
+++ b/app/models/event_instance.rb
@@ -64,6 +64,12 @@ class EventInstance < ApplicationRecord
     yt_video_id && "https://youtu.be/#{yt_video_id}"
   end
 
+  def channels_for_event
+    return [] if event_id.nil?
+    event = Event.find(event_id)
+    event.slack_channel_codes
+  end
+
   private
 
   def manually_updated_event_not_finished?
--- a/app/services/hangout_notification_service.rb
+++ b/app/services/hangout_notification_service.rb
@@ -22,6 +22,7 @@ class HangoutNotificationService
     return if @event_instance.hangout_url.blank?
     
     channels = channels_for_project @event_instance.project
+    channels += channels_for_event @event_instance.event_id
     message = "#{@event_instance.title}: <#{@event_instance.hangout_url}|click to join>"
     @here_message = "@here #{message}"
     @channel_message = "@channel #{message}"
@@ -33,6 +34,13 @@ class HangoutNotificationService
     return [] if project.nil? or project.slug.nil?
     project.slack_channel_codes
   end
+
+  def channels_for_event event_id
+    return [] if event_id.nil?
+    event = Event.find(event_id)
+    event.slack_channel_codes
+  end
+
   
   def send_notifications channels
     return post_premium_mob_hangout_notification if @event_instance.for == 'Premium Mob Members'
--- a/app/services/hangout_notification_service.rb
+++ b/app/services/hangout_notification_service.rb
@@ -22,7 +22,7 @@ class HangoutNotificationService
     return if @event_instance.hangout_url.blank?
     
     channels = channels_for_project @event_instance.project
-    channels += channels_for_event @event_instance.event_id
+    channels += @event_instance.channels_for_event @event_instance.event_id
     message = "#{@event_instance.title}: <#{@event_instance.hangout_url}|click to join>"
     @here_message = "@here #{message}"
     @channel_message = "@channel #{message}"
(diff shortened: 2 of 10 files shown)


tansaku:
looks like a great start @mattwr18 - we had a partial start on this in https://github.com/AgileVentures/WebsiteOne/pull/3077 which I assume inspired at least some of the direction for this, but no problem if not.

It's got a bit late on a Friday for me to really get my mind into this.  Since we don't have event - channel relations I don't think we need a rake task - in a first pass they can be added in rails console I think for the few we need, and then further updated once we push out a UI change to allow live editing of them ...

Without having looked at it too closely I'd say let's get it out and create refactoring tickets for any concerns - please feel free to merge if you want to deploy next week.

mattwr18:
Actually, I pretty much copied and pasted the code from that `PR`! then it was just a matter of updating the notification services specs and getting them passing. 

I think I might actually merge it, since you are ok with it

app/services/youtube_notification_service.rb
```diff
@@ -31,6 +32,12 @@ def channels_for_project project
     project.slack_channel_codes
   end
 
+  def channels_for_event event_id
+    return [] if event_id.nil?
+    event = Event.find(event_id)
+    event.slack_channel_codes
+  end
+
```
NickSchimek: hmmm this bit of code seems like it could live in the Event model?

app/services/hangout_notification_service.rb
```diff
@@ -33,6 +34,13 @@ def channels_for_project project
     return [] if project.nil? or project.slug.nil?
     project.slack_channel_codes
   end
+
+  def channels_for_event event_id
+    return [] if event_id.nil?
+    event = Event.find(event_id)
+    event.slack_channel_codes
+  end
+
```
NickSchimek: hmmm should this be in the Event model?

app/services/youtube_notification_service.rb
mattwr18: this method `slack_channel_codes` is in the Event model...
it calls `slack_channels.pluck(:code)`

mattwr18: or do you mean the entire method `channel_for_event`?

NickSchimek: Yes, I mean the entire method. The method calls the Event model and has an `event id` parameter so it seems that it should be in the Event model or perhaps event instance....

NickSchimek: Also if the method is in the model then the duplication in `hangout_notification_service.rb` can be removed.

mattwr18: yes, great advice! I was saying to @gr1d99 that maybe we should be extracting into a helper method, but I like the idea of add it to the Event model

we could do something similar with
```
def channels_for_project project
    return [] if project.nil? or project.slug.nil?
    project.slack_channel_codes
  end
```
maybe create a separate ticket to refactor it.

lib/channels_list.rb
```diff
@@ -58,7 +58,8 @@ module ChannelsList
       'pairing_notifications': 'C29J3DPGW',
       'standup_notifications': 'C29JE6HGR',
       'premium_extra': 'C29J4QQ9M',
-      'premium_mob_trialists': 'C29J4QQ9F'
+      'premium_mob_trialists': 'C29J4QQ9F',
+      'new-members': 'C02G8J699'
```
NickSchimek: Is there a plan to move these to the database?

mattwr18: yes! 
we have already put in a `PR` that has been merged to `master`, which uses the projects slack channels join table... we create a rake task to go through this hard coded channels to the database.
After this has made its way to production, I will be adding some channels manually to the events slack channels join table, then I'll put in a third `PR` to allow users to add the slack_channels to the event themselves through the interface.

mattwr18:
@NickSchimek, I followed your advice and think it's now improved, could you have a look?

app/models/event_instance.rb
```diff
@@ -64,6 +64,12 @@ def yt_url
     yt_video_id && "https://youtu.be/#{yt_video_id}"
   end
 
+  def channels_for_event event_id
+    return [] if event_id.nil?
+    event = Event.find(event_id)
+    event.slack_channel_codes
+  end
```
NickSchimek: The event_id parameter is not required as it's an attribute of event_instance. 
I'd suggest updating the method to this using `self` to get the attribute of the instance.
```
  def channels_for_event
    return [] if self.event_id.nil?
    event = Event.find(self.event_id)
    event.slack_channel_codes
  end
```

Then the method can be called from the hangout/youtube notification service method without passing an argument. Example: `@event_instance.channels_for_event`

mattwr18: that's great! I was thinking it would be nice to not have to pass it in! I'll update it and push it up :smile_cat:

NickSchimek: As in option I believe in Ruby you can do `attr_reader :event_id` and then you don't have to use the `self` keyword and can just say `event_id`. Whichever way works for you.

NickSchimek: You may need to use `attr_reader :event_id` regardless. I'm not 100% sure.

NickSchimek:
Nice. Looks great @mattwr18 !

mattwr18:
@tansaku, I didn't end up getting this merged... it's not really a rush, so if you want to have a look and merge maybe? 
I prefer not to merge my own code unless there isn't a project manager or team

tansaku:
great work @mattwr18 and really appreciate the input from you @NickSchimek - let's get this deployed! :-)

PR #3187, opened 2019-05-15
Updates ruby version to 2.5.1 in Docker file

abdellani:
#### Issue addressed

fixes #


mattwr18:
thanks @abdellani, we've mean to do this a while back :smile_cat:

PR #3192, opened 2019-05-16
149 adjust passwords endpoint

bryszard:
Adjusting the password endpoint to handle reset_password POST and PUT requests from the frontend app. On WebsiteOne-FE - https://github.com/AgileVentures/WebsiteOne-FE/issues/149


tansaku:
thanks for this @bryszard - seems like it's failing based on ruby versions ...

```
Your Ruby version is 2.5.1, but your Gemfile specified 2.5.3
```

mattwr18:
@tansaku, I don't know if you noticed, but this `PR` is to be merged into `dokku_deploy_branch`, which is using `Ruby 2.5.3` because of the official ruby docker image and its automatic updates.

I will change this though as Semaphore is not adding new Ruby versions so quickly.

tansaku:
thanks @mattwr18 :-)

PR #3039, opened 2019-02-06
Check that languages from multiple repos are imported

mattwr18:
Pulls in languages from all a project's repos
#### Issue addressed

fixes #2725 

#### Testing

more than one github_url add to a project's source_repositories, check that languages from the 2nd or 3rd repo appear in a project's languages

PR #3198, opened 2019-05-20
Update github languages job, fixture

### LINKED ISSUE #3197: Build failing on github languages cuke
why this started failing right now is unclear as it doesn't seem related to any recently merged changes, but it is reliably failing locally in the same way

Screenshot at 2019-05-20 18:22:31

## Expected Behavior

this shouldn't be blocking our deploys
## Current Behavior

it is blocking our deploy pipeline because it is failing on the `develop` branch and `staging`
## Possible Solution

it seems due to some intermittent behavior of `find_or_create_by`
https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

mattwr18:
- on unique constraint error, retry, apparently due to flakiness of `find_or_create_by` https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by
- not sure what the diff between this fixture is, it can be removed if it's not needed to be updated

#### Issue addressed

fixes #3197

### DIFF
--- a/app/jobs/github_languages_job.rb
+++ b/app/jobs/github_languages_job.rb
@@ -24,7 +24,8 @@ module GithubLanguagesJob
   def add_new_languages_to(project)
     begin
       new_languages_for(project).each { |language| project.languages << Language.find_or_create_by(name: language) }
-    rescue ActiveRecord::RecordNotUnique
+    rescue ActiveRecord::RecordNotUnique => error
+      log_error(error, project)
       retry
     end
   end
--- a/app/jobs/github_languages_job.rb
+++ b/app/jobs/github_languages_job.rb
@@ -22,7 +22,11 @@ module GithubLanguagesJob
   end
   
   def add_new_languages_to(project)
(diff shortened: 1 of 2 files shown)


mattwr18:
it appears that the fixture was not updated when the `PR` to check that languages from multiple repos was merged in, see https://github.com/AgileVentures/WebsiteOne/pull/3039/files

@tansaku this issue was causing the build to fail consistently and on exploration, starting failing consistently for me locally on the same cuke. As I explained in the issue it seems due to the flaky nature of our use of `find_or_create_by`... found the solution to the issue here https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

Instead of merging this in myself, I thought I'd leave it to you to review and merge if you are happy with it, then I can continue the deploy to production.

tansaku:
great work @mattwr18 tracking this down - looks good to me :-)

it might be worth adding a logging statement before retry just so there's a trace if it gets stuck in a loop in future maybe?

tansaku:
great work @mattwr18 - big thanks for this :-)

Issue #2926, opened 2018-12-10
Add validation for repeat_ends for the Event model


NickSchimek:
## Description
When an event is created there are a couple different options based on the type of event, however in either case whether its a recurring event or a non-recurring event `repeat_ends` should be set regardless. 

When an event is created through the UI there are no issues and repeat_ends is set accordingly. It's set to `true` for non-recurring events and repeating events with an end date. It is set to `false` for repeating events without an end date.

### Steps to fix:
- [ ] Add validation for repeat_ends field
- [ ] Update any tests (if any) that break

Danielshow:
hey @NickSchimek, I can work on this.

NickSchimek:
Thanks @Danielshow Go for it! Let me know if there are any questions.

Danielshow:
Done  @NickSchimek, I have raised a PR for this

PR #3211, opened 2019-05-28
Add validation for repeat_ends in event model

Danielshow:
#### Issue addressed
This PR fixes the issue below by validating repeat_ends field in Event model

fixes [#2926](https://github.com/AgileVentures/WebsiteOne/issues/2926)

#### Testing
Fix failing test by adding repeat_ends field to factories

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -16,6 +16,7 @@ class Event < ApplicationRecord
   validates :repeats_every_n_weeks, :presence => true, :if => lambda { |e| e.repeats == 'weekly' or e.repeats == 'biweekly' }
   validates :repeat_ends_on, :presence => true, :allow_blank => false, :if => lambda{ |e| (e.repeats == 'weekly' or e.repeats == 'biweekly') and e.repeat_ends_string == 'on' }
   validate :must_have_at_least_one_repeats_weekly_each_days_of_the_week, :if => lambda { |e| e.repeats == 'weekly' or e.repeats == 'biweekly' }
+  validates :repeat_ends, inclusion: { in: [true, false]  } 
   attr_accessor :next_occurrence_time_attr
   attr_accessor :repeat_ends_string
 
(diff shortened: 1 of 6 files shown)


tansaku:
@Danielshow this looks great!

I guess my only question would be whether we need some new test to expose the issue being fixed?  what do you think?

Danielshow:
Sure @tansaku, I will add a test to the models to make sure the field is always added. Will update soon

NickSchimek:
Thanks @Danielshow for the work you put into this. It looks like great work to me!

@tansaku The issue #2926 probably should’ve been closed or redefined after fixing the Event query. I definitely overlooked that.

After I wrote the issue I had done more investigation and I realized the main issue was that the `repeat_ends_on` string is incorrect when an event is updated from a `recurring event with an end date` to a `recurring event that never ends`, because the `repeat_ends_on` field does not get changed in the front end and sends up the old information to the backend on update. It’s actually a front end issue, not a backend issue. Realizing this I was able to fix the Event query without having to change anything in the front end, but inadvertently left the ticket open without updating it.

It's actually not a validation issue, but an issue with the information coming from the front end. This can be seen if you open up the network tab and watch the params being sent. By creating an event that ends on 5/31, and then editing the event to never ends the front end and verifying it still sends up the ends on date 5/31 as seen here:

Screenshot 2019-05-29 12 09 19

Those two pieces of information in an event object contradicts each other.

Not sure what to do here. This is great work you put into this PR, but I would also like the repeat ends on string fixed. What do you think @Danielshow ? I guess I could update the old issue or create a new issue. Let me know how you feel about it.

Danielshow:
Sure, @NickSchimek. You can update or create a new issue and assign it to me. I will love to work on it

tansaku:
great work @NickSchimek and @Danielshow - I think if we have this green with a test let's merge and create a new issue for anything outstanding - the sooner we can get things into production the better I say :-)

I'll merge tomorrow unless anyone objects ...

NickSchimek:
@Danielshow issue has been created #3215 if you'd like to work on it just make a comment stating that your working on it as I wasn't able to assign you to it. Thanks for all your help!

PR #3216, opened 2019-05-31
Persist empty string when repeat end is set to never

### LINKED ISSUE #3215: Persist empty string when repeat ends is never
## Description
If a user sets `repeat ends` to `on` and selects an `end date` in the UI, and then changes the `repeats ends` to `never` the frontend-end still sends the `end date`, and it is persisted when it shouldn't be.

## Expected Behavior

The expected behavior is that when a recurring event never ends then it shouldn't have a repeat ends date persisted.

## Current Behavior

here is a screen cap of the params being passed when changed on the UI.
Screenshot 2019-05-29 12 09 19

## Possible Solution

I think this should be changed in the front-end so that when never is selected it sends an empty string or whatever the current default is (nil?).

1. set repeats ends to on
2. set a date
3. set repeat ends to never
4. submit

Danielshow:
#### Issue addressed
Fix UI bug when repeat end is set to never on the UI

[Fixes #3215](https://github.com/AgileVentures/WebsiteOne/issues/3215)

#### Testing
N/A

### DIFF
--- a/app/assets/javascripts/event_datepicker.js
+++ b/app/assets/javascripts/event_datepicker.js
@@ -31,7 +31,7 @@ var eventDatepicker = {
         switch ($('#event_repeat_ends_string').val()) {
             case 'never':
                 $('#repeat_ends_on_label').hide();
-                $('#repeat_ends_on').hide();
+                $('#repeat_ends_on').hide().val('');
                 break;
             case 'on':
                 $('#repeat_ends_on_label').show();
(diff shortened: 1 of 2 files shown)


tansaku:
great work @Danielshow - do we need a test to expose the bug as well?

Danielshow:
Hey @tansaku, that will be an e2e test. I will check that out.

Danielshow:
Done @tansaku, I think it is good to merge

tansaku:
@Danielshow great - looks like you added a front end js unit test - which is great - if @NickSchimek is happy with that and that we don't need any changes to the cukes, then I'm happy to have it merged in - really appreciate your work on this :-)

Issue #2968, opened 2018-12-27
Remove destroy action in projects controller


NickSchimek:
## Description
The Destroy action in the Projects controller is empty. We currently do not allow user's to destroy projects. This should be removed along with the corresponding route.

This is done when the following tasks are completed...
- [ ] Update the `config/routes` to remove the DELETE route from the Projects resource.
- [ ] Remove the destroy action in the Projects controller.
- [ ] If a view template exists then remove that too.

Danielshow:
Hey @NickSchimek, this has not been fixed.  I can pick it up

NickSchimek:
Yep, go for it! Thanks!

Danielshow:
Done @NickSchimek

PR #3228, opened 2019-06-09
Remove destroy action from Project Controller

Danielshow:
## Description
This pull request removes destroy action from projects controller since it is not been used.
 - Delete commented test
-  exempt destroy routes

#### Issue addressed

fixes #[2968](https://github.com/AgileVentures/WebsiteOne/issues/2968)

N/A

#### Testing
N/A

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,7 +1,7 @@
 class ProjectsController < ApplicationController
   layout 'with_sidebar'
   before_action :authenticate_user!, except: [:index, :show]
-  before_action :set_project, only: [:show, :edit, :update, :destroy]
+  before_action :set_project, only: [:show, :edit, :update]
   before_action :get_current_stories, only: [:show]
   include DocumentsHelper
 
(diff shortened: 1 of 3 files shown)


NickSchimek:
Thanks for doing this @Danielshow . Could we also remove the `before_action` hook for the `:destroy` method on line 4? https://github.com/AgileVentures/WebsiteOne/blob/50e1d04d22b75354330504756abc1b80bb6231c5/app/controllers/projects_controller.rb#L4

Also on the bottom of the file it white lists some parameters. I see there is a `:_destroy`, I'm not sure if we can also remove those two references too. Do you have any insight on what that is referring to, and if we can safely remove them?
https://github.com/AgileVentures/WebsiteOne/blob/50e1d04d22b75354330504756abc1b80bb6231c5/app/controllers/projects_controller.rb#L129
https://github.com/AgileVentures/WebsiteOne/blob/50e1d04d22b75354330504756abc1b80bb6231c5/app/controllers/projects_controller.rb#L130

Danielshow:
I will update now but for line 129 and 130, I don't have an idea of what it is doing. I will check it out though

Danielshow:
Hey @NickSchimek, the project model accepts nested attributes `source_repositories` and `issue_tracker`. I guess the `_destroy` keyword in the params is used when checking a box to delete any of the parameter entered for the nested attributes.

NickSchimek:
I understand. Thanks for looking into that.

PR #3265, opened 2019-06-25
Wrapped LOWER sql call with Arel.sql

### LINKED ISSUE #2924: Deprecation warning in tests.
## Description
Seeing the following deprecation warning when running cucumber tests.

DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "LOWER(title)". Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from search at /home/runner/WebsiteOne/app/models/project.rb:39)

Tasks: - Update the following query method.
- [ ] https://github.com/AgileVentures/WebsiteOne/blob/develop/app/models/project.rb#L39

clementkng:
Removes the `Dangerous query method` DEPRECATION warning when running cucumber tests. 
#### Issue addressed

fixes #2924

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -39,7 +39,7 @@ class Project < ApplicationRecord
   end
 
   def self.search(search, page)
-    order('LOWER(title)')
+    order(Arel.sql('LOWER(title)'))
       .where('title LIKE ?', "%#{search}%")
       .paginate(per_page: 9, page: page)
   end


tansaku:
big thanks for this @clementkng ! :-)

PR #3278, opened 2019-07-02
Set Codeclimate duplicate and similar code threshold to 3

### LINKED ISSUE #2976: adjust code climate to only flag duplication of 3 rather than 2

clementkng:
Updated to [version 2](https://docs.codeclimate.com/docs/advanced-configuration#section-analysis-configuration-versions) used in Codeclimate, which allows tunable duplicate code checks.
WIP b/c I'm not sure how to test this w/o an open source project and need to tune the exact values of various `threshold` constants.

#### Issue addressed

fixes #2976

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,5 +1,12 @@
 ---
 version: "2"
+checks:
+  similar-code:
+    enabled: true
+    config:
+      threshold: 3
+      languages:
+      - ruby
 plugins:
   brakeman:
     enabled: true
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,41 +1,38 @@
 ---
-engines:
+version: "2"
+checks:
+  similar-code:
+    enabled: true
+    config:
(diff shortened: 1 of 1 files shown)


clementkng:
I'm also not sure if the old duplication warnings can be cleared, but using [codeclimate-duplication](https://github.com/codeclimate/codeclimate-duplication) should mean that new duplicate code raises warnings on 3 instances and not 2.

clementkng:
I'm also not sure if the old duplication warnings can be cleared, but using [codeclimate-duplication](https://github.com/codeclimate/codeclimate-duplication) should mean that new duplicate code raises warnings on 3 instances and not 2.

tansaku:
this looks great - thanks so much @clementkng

clementkng:
@tansaku One concern I may have is that I'm not sure if this messes w/ different default thresholds for codeclimate  ie similarity measures for different languages and the other measures of code health. We should revisit this if the standards are noticeably too lax.

PR #3367, opened 2019-09-12
Refactor hire me action

### LINKED ISSUE #3274: Refactor hire me form action in users controller
## Description

As part of the refactoring mob we will be refactoring this method in our weekly sessions:
https://github.com/AgileVentures/WebsiteOne/blob/develop/app/controllers/users_controller.rb#L28

NickSchimek:
## Description
Refactored the Hire Me action that is responsible for sending an email to a member when a user would like to hire the member.

Created a contact object to handle the validations, however since this information is not persisted we used the Form object pattern as suggested by @ciremoussadia 

Here is more information on the form object pattern.

https://medium.com/@jaryl/disciplined-rails-form-object-techniques-patterns-part-1-23cfffcaf429

https://thoughtbot.com/blog/activemodel-form-objects

Our desire to refactor was due to the complexity of the code as the method was long and hard to understand. CodeClimate now reports that two outstanding issues have been fixed.

The behavior has slightly changed. Now when the form has validation errors the entered data is not lost when rendering the show page and when the user clicks on the form again it is already filled out with the information that it had previously.

The controller was capturing the `request.env[HTTP_REFERER]` to see where the request was coming from and attempted to always redirect `back` to this page. However the contact form is only available from the users show page, and so it was updated to always redirect back to the uers show page.

Big thanks to @cesswairimu and @ciremoussadia for helping with this refactoring!!!

fixes #3274

#### Testing
Some tests in the users controller were moved from the controller to the model as they were testing validations. Two tests were simply deleted as they were trying to test the system, and those tests already existed as feature specs. Error messages were updated in the tests, because we now use the default validation error messages provided by active record.


aonomike:
This looks really great @NickSchimek @cesswairimu @ciremoussadia, I should not be missing the refactoting mob really!!!

spec/controllers/users_controller_spec.rb
```diff
@@ -105,93 +127,43 @@
...
-        post :hire_me_contact_form, params: { message_form: {} }
-        expect(flash[:alert]).to eq 'Please fill in Name, Email and Message field'
-      end
-
-      it 'should not fail with no back path' do
+      it 'should fail with no back path' do
         request.env['HTTP_REFERER'] = nil
-        post :hire_me_contact_form, params: { message_form: { name: '', email: '', message: '' } }
-        expect(flash[:alert]).to eq 'Please fill in Name, Email and Message field'
+        post :hire_me, params: empty_params
```
aonomike: much more readable, :)

ciremoussadia: @aonomike u'll enjoy it.

aonomike:
@NickSchimek I will just go ahead and merge this as it works for me locally.

PR #3414, opened 2019-10-17
Refactoring subscription controller

### LINKED ISSUE #3002: subscription controller name refactoring
we noticed that some of the variable names in the subscription code are a bit confusing - check out the subscriptions controller:

```rb
  def new
    @plan = detect_plan_before_payment
  end

  def create
    @user = detect_user
    @plan = detect_plan_after_payment
    @sponsored_user = sponsored_user?
```

here the `@upgrade_user` contains a slug that is only present in the case of sponsorship, so would be better named maybe `sponsee_slug`

`@user` is in fact an actual user, but only present in the case of sponsorship, so would be better as `@sponsee` and `@sponsored_user` is a boolean - could be confused for a user - maybe better to rename as `is_sponsorship` - similarly for the method name

maybe some of these methods/vars that are specific to the views, should be moved from the controller to the view helper ...

snpd25:
fixes #3002

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -5,19 +5,19 @@ class SubscriptionsController < ApplicationController
   # skip_before_action :verify_authenticity_token, only: [:create], if: :paypal?
 
   def new
-    @upgrade_user = params[:user_id]
-    @sponsorship = @upgrade_user && current_user.try(:id) != @upgrade_user
+    @sponsee_slug = params[:user_id]
+    @sponsorship = @sponsee_slug && current_user.try(:id) != @sponsee_slug
     @plan = detect_plan_before_payment
   end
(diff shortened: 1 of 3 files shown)


snpd25:
@aonomike , please review this.

aonomike:
Hi @snpd25 thanks so much for working in this 👍 Have you seen that we are failing cucumber tests on CI? have you maybe tried to run the specific failing tests locally to see if they pass? We can plan a pair programming session just incase you want to :)

aonomike:
Hi @snpd25 many thanks for fixing the tests, maybe just another 'silly' request from my end, do you mind branching off your develop to a branch, maybe named as `3002_refactor-subscription-controller`. In this case I dont get so much confused with local branches that I pull from your fork while reviewing. Great job so far!

PR #3422, opened 2019-10-22
Refactoring subscription controller

### LINKED ISSUE #3002: subscription controller name refactoring

snpd25:
fixes #3002 
Refer #3414

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -5,19 +5,19 @@ class SubscriptionsController < ApplicationController
   # skip_before_action :verify_authenticity_token, only: [:create], if: :paypal?
 
   def new
-    @upgrade_user = params[:user_id]
-    @sponsorship = @upgrade_user && current_user.try(:id) != @upgrade_user
+    @sponsee_slug = params[:user_id]
+    @sponsorship = @sponsee_slug && current_user.try(:id) != @sponsee_slug
     @plan = detect_plan_before_payment
   end
 
   def create
-    @user = detect_user
+    @sponsee = detect_user
     @plan = detect_plan_after_payment
-    @sponsored_user = sponsored_user?
+    @is_sponsorship = is_sponsorship?
     
     create_stripe_customer unless paypal?
 
-    add_appropriate_subscription(@user, current_user)
+    add_appropriate_subscription(@sponsee, current_user)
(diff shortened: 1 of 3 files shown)


snpd25:
> maybe we could refactor the name of @sponsorship as well? its a boolean that checks that the sponsee is not the current user. Maybe we can rename it to @sponsee_is_not_current_user?
> Also, current_user.try(:id) returns the id (integer) that we are comparing with @sponsee_slug, this will always be false, are we sure we are comparing the correct thing or do we want to compare current_user.try(:slug) with the sponsee slug 🤔 ? I know maybe this wasnt the scope of this PR but maybe we need to see if that's a bug?

I think for the time being, since these changes are needed and approved, we should keep them. I will try to see if that is a bug and if required will send another PR for that. We can discuss the changes there :)

aonomike:
> > maybe we could refactor the name of @sponsorship as well? its a boolean that checks that the sponsee is not the current user. Maybe we can rename it to @sponsee_is_not_current_user?
> > Also, current_user.try(:id) returns the id (integer) that we are comparing with @sponsee_slug, this will always be false, are we sure we are comparing the correct thing or do we want to compare current_user.try(:slug) with the sponsee slug 🤔 ? I know maybe this wasnt the scope of this PR but maybe we need to see if that's a bug?
> 
> I think for the time being, since these changes are needed and approved, we should keep them. I will try to see if that is a bug and if required will send another PR for that. We can discuss the changes there :)

I have [created an issue](https://github.com/AgileVentures/WebsiteOne/issues/3427) as I think we may need to write a test that fails and then fix it to make pass just incase the case is true. Thanks  for working on this!!

PR #3528, opened 2020-04-19
Maintenance

NickSchimek:
Did some maintenance to help keep the repo updated.
- Rename method

filter is a core ruby method, and this name was clashing with the core
ruby method.

- Update RuboCop cop names
- Performance cops have been moved to seperate gem.
- Require rubocop performance
- Require rubocop rails
- Update Faker deprecated arguments

I'm going to hold off on updating the Ruby version until we have finished migrating to semaphore v2.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -24,6 +24,7 @@ plugins:
     enabled: true
   rubocop:
     enabled: true
+    channel: rubocop-0-77
 exclude_patterns:
 - config/
 - db/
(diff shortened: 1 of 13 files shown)


NickSchimek:
@aonomike In order to get semaphoreci to run the build. We will need to update the ruby version. It seems for versions above 2.5.1 we will have to explicitly require it in the build command mentioned here: https://semaphoreci.com/docs/languages/ruby/ruby-support-on-semaphore.html#installing-additional-version-of-ruby

It appears I don't have admin access for semaphoreci. Is this something you want to do?

aonomike:
> we

Thanks for this one @NickSchimek, let me see if I have admin access for semaphore.

PR #3666, opened 2021-03-30
Update mission_statement.md

Georgy5:
### What
remove waffle.io link
### Why
Waffle.io closed down in 2019.
### Source
https://projectmanagernews.com/general/what-happened-to-waffle-io/


aonomike (approved):
Thanks @Georgy5 for this.

Issue #3778, opened 2021-07-17
Create Project approval flow


tochman:
A new project should be approved by an admin before being displayed on the site

```
As a Site Administrator
In order to keep the Projects list free from spam
I would like to be able to make a decision on a  new project prior to making it public
```

wordhouse66:
I will try using projects' :status so when a project is created it will be status: 'Pending' by default and then you will have a page of pending projects and you will be able to click a button to change the status to 'Active' or delete the project. ONLY ACTIVE PROJECTS WILL BE VISIBLE ON THE PROJECTS INDEX PAGE. And only admin will access the pending projects page. How does that sound?

wordhouse66:
I think I should add two pages for the user so when they create a project they have 'my_pending_projects' page as well as 'my_active_projects' page. That way they can see their projects even when it's not yet approved and they don't have to wonder where it went. The text in the project show page will state that "this project is currently under review and will be made public only when approved by Agile Ventures admin"

mattlindsey:
Given I am a user
When I create a project
Then I should get an email notification to that effect and that an admin will look at it soon. 

And I should also receive an email informing me of the Admin's approval of my project if it is indeed approved.

At the moment this does not happen and the user is left to wonder what is going on with the project he or she created.

PR #3784, opened 2021-07-19
Admin based Project approval flow 

wordhouse66:
Create Project approval flow #3778

#### Issue addressed
fixes issue #3778  in https://github.com/AgileVentures/WebsiteOne/issues/3778

To fix the issue I have created methods in the projects controller to either activate or deactivate a project. The methods are only accessible to the admin who also is the only one who can see the corresponding buttons in the project show page that triggers the actions. 
I also modified the index method to specifically have @projects refer to only those with (status: "Active"). 
A new page and method called pending_projects is created to display the projects with (status: "pending") . This method is also only accessible to the user for whom (user.admin? returns true).
I also added a method call access_edit to restrict the access to a project edit page to only the creator and admin. however only the admin can change the status to "Active". 
Screenshot 2021-07-19 at 08 03 06

There are also corresponding routes for all the new methods created in routes.rb.


config/routes.rb
```diff
@@ -73,6 +76,11 @@ def loaderio_token
     end
   end
 
+  # get '/pending_projects' => 'projects#pending_projects'
+  # post '/activate_project/:id', controller: :projects, action: :activate_project
+  # post '/activate_project/:id' => 'projects#activate_project', as: 'activate_project'
+  # post '/deactivate_project/:id' => 'projects#deactivate_project', as: 'deactivate_project'
```
tochman, line 82: De faq is this?

Issue #3790, opened 2021-07-30
gemfile expects ruby 3 but docker image is 2:7


mtc2013:
Maybe it needs updating?

mtc2013:
Fixed by #3802

PR #3802, opened 2022-10-02
Fix docker build

mtc2013:
This fixes the docker build and moves to ruby 3.0.0.  The need to move to ruby 3.0.0 was requested here https://github.com/AgileVentures/WebsiteOne/issues/3790 .  The Docker automated shell script, setup.sh, was failing to build the project.  This was tested on Linux.  

Basically, we had to make a change or two to the dockerfile itself and then to the database.yml in Rails.  This should continue to work the same for others who don't use Docker.

And now setup.sh and start.sh function as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM ruby:2.6.3
+FROM ruby:3.0.0
 
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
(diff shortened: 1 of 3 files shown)


yosefbennywidyo (approved):

mattlindsey (approved):
Looks good. I see that you update ruby to 3.0.4 in the next PR, so I'll look at that now too.

mtc2013:
Fix issue #3790

PR #3814, opened 2022-12-13
Replace jasmine gem and phantomjs

### LINKED ISSUE #3805: Replace no longer maintained phantomjs
The phantomjs headless browser for javascript testing stopped being maintained 5 years ago and should be replaced.  Maybe headless chrome.

Found these:
https://www.bigbinary.com/blog/replacing-phantomjs-with-headless-chrome
https://medium.com/@vermaxik/how-to-replace-phantomjs-poltergeist-with-cuprite-gem-a656da626430

mattlindsey:
Remove phantomjs from cucumber and rspec tests and replace with the cuprite gem using headless chrome. I tried first using selenium (rather than cuprite), but that seemed to have many more issues.
Some tests were commented out due errors in cuprite, but can be added back later if deemed useful. I also commented out Bootstrap tooltips due to errors in cuprite, but these could be added back for non-test environments with a conditional statement, if desired.

Note that I had to update chromedriver on my mac by doing 'brew update chromedriver.'

fixes #3805 
fixes #3806


mtc2013 (approved):
I think this is good as it's going to be for now.  We can merge this in.

PR #3815, opened 2022-12-13
WIP upgrade rails gems to 6.1.7

### LINKED ISSUE #3810: upgrade to latest rails 6.1 version
We should ride the latest version of 6

mattlindsey:
fixes #3810


mattlindsey:
I think that I successfully updated event_instances.exclusions to serialize using JSON instead of YAML, but event_instances.participants is broken.  I left some attempt to fix it in:

rspec ./spec/presenters/event_instance_presenter_spec.rb

mattlindsey:
Don't apply that last commit.

I think the problem is the 2nd migration I did with the drop and add column doesn’t work.  I was supposed to use ‘change_column.’  The 1st migration worked because initially I called it a different name.  I need to revert and redo both using ‘change_column.’

Issue #3819, opened 2022-12-18
Upgrade to rails 7


mattlindsey:
Upgrade to rails 7.

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html

Florence-Njeri:
@mattlindsey I'd love to help with the Rails upgrade

mattlindsey:
Hi @Florence-Njeri.  That's great! Let me know if you have questions about the setup. We have a chat on slack that I can send you a link for.

Florence-Njeri:
Hi Matt,

I’m currently setting up the developer environment and I require access to
the .env file content. I’d also appreciate it if you sent me the slack
invite.

Best regards,
Florence Njeri.

mattlindsey:
Great.  I think you can get an invite here: https://agileventures.slack.com/messages/websiteone and I can give the the .env file there.
Matt

Florence-Njeri:
There are only the register/login options. The register options limit me to
use these domains:

   - agileventures.org
   - berkeley.edu
   - binghamton.edu

mattlindsey:
@Florence-Njeri . Oh, sorry. Usually you would get an email, but can probably use this: https://join.slack.com/t/agileventures/shared_invite/zt-1587bzqk7-Y8ACcBKmwi0SWbI3Iu78eg

mattlindsey:
@Florence-Njeri.  Someone just posted this, which looks good: https://github.com/railslink/resources/wiki/Rails:-Upgrade-process
We recently upgraded to latest rail 6, so maybe 7 won't be too painful.

Florence-Njeri:
Thanks, I'll check it out.

mattlindsey:
@Florence-Njeri.  No rush, but are you still working on this?

Florence-Njeri:
Hi Matt,

This task is in progress. Hoping to be done by end of the week.

mattlindsey:
@Florence-Njeri.  Great, thanks.  Don't forget to do 'git pull upstream develop' to get recent updates to the develop branch.

mattlindsey:
I think you can probably go right to version 7.0.4.2

Issue #3813, opened 2022-12-12
Restore saving VCR cassettes in github


mattlindsey:
We should consider restoring the saving of VCR cassettes in github ,which was removed recently.

It’s easier to develop and manage without them, but then there’s the potential of having a problem, where after a failure it messes up subsequent runs that should be marked success.

PR #3821, opened 2022-12-20
Add support for google cloud

mattlindsey:
Modifications to allow running the site on docker locally and on Google Cloud.

It's currently dying on trying to connect to Google Cloud SQL postgres with this:
  PG::ConnectionBad: could not connect to server: No such file or directory
  
It seems to happen first when 'rake db:setup' tries to create the db and do migrations.  And the postgres logs don't show it hitting the database.  And the 'websiteone-production' db doesn't get created unless I do it manually.

I have checked a lot of things, and it may be a problem with the encrypted db password using the 'master key' method in rails, but I also seem to get the error just defining the password in the cloud service.  Possibly the master key or 'rails secret key' didn't get saved correctly.


mtc2013:
have a look at this for sql issues

https://dev.to/ku6ryo/access-cloud-sql-from-docker-on-cloud-run-41b6

mattlindsey:
Also this for postgres stuff needed
https://dev.to/ku6ryo/access-cloud-sql-from-docker-on-cloud-run-41b6

mattlindsey:
I think this looks good now.
Turns out this doc was the latest/best way to do it, rather than the older 4 step tutorial which is more complicated:
https://cloud.google.com/ruby/rails/run

.gitignore
```diff
@@ -102,4 +102,5 @@ certbot.log
 rerun.txt
 
 # Ignore VCR
-features/support/fixtures/cassettes/**
\ No newline at end of file
+features/support/fixtures/cassettes/**
```
mtc2013, line 105: is it necessary to put this re-enabling of vcr in this pr?

Gemfile
```diff
@@ -36,6 +36,7 @@ gem 'faker'
 gem 'font-awesome-rails'
 gem 'friendly_id'
 gem 'geocoder'
+gem 'google-cloud-storage'
```
mtc2013: do we need this in this pr?

config/application.rb
```diff
@@ -2,6 +2,7 @@
 
 require_relative 'boot'
 require 'rails'
+require "active_storage/engine"
```
mtc2013: do we end up needing this?

config/storage.yml
```diff
@@ -15,11 +15,12 @@ local:
 #   bucket: your_own_bucket
 
 # Remember not to checkin your GCS keyfile to a repository
-# google:
-#   service: GCS
-#   project: your_project
-#   credentials: <%= Rails.root.join("path/to/gcs.keyfile") %>
-#   bucket: your_own_bucket
+# [START cloudrun_rails_storage]
+google:
```
mtc2013: do we need this in here?

db/migrate/20221221225406_create_active_storage_tables.active_storage.rb
```diff
@@ -0,0 +1,48 @@
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
```
mtc2013: maybe we don't need this in this pr?

Dockerfile
```diff
@@ -1,41 +1,66 @@
...
 
-RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
-RUN apt-get update -qq && apt-get install -y build-essential \
-    libpq-dev nodejs  \
-    libqt5webkit5-dev dos2unix \
-    gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x
+# [START cloudrun_rails_base_image]
+# Pinning the OS to buster because the nodejs install script is buster-specific.
+# Be sure to update the nodejs install command if the base image OS is updated.
+FROM ruby:3.0-buster as base
```
mtc2013, line 7: are we running 3.0.5 here with this?

.gitignore
mattlindsey: Not necessary for google cloud, but leaving them disabled can cause confusion in development (as it did with me).  I don't think the PR that took them out gave any reason for taking it out, so I'd rather keep it the way it has always been unless there's a good reason.

Gemfile
mattlindsey: I'll try taking all of the cloud storage stuff out.  There's a chance that it's necessary for asset pipeline, but hopefully not.

Dockerfile
mattlindsey: Yes.  I think they keep 'buster' up with the 'stable' version of ruby 3.0.x

mattlindsey:
fixes #3819 
fixes #3813

mattlindsey:
Based on the PR review comments, I tried to take out the Google Storage stuff but it seems that at least some of it is necessary, since we go back to getting this error:

Step #2 - “apply migrations”: PG::ConnectionBad: could not connect to server: No such file or directory
Step #2 - “apply migrations”: 	Is the server running locally and accepting
Step #2 - “apply migrations”: 	connections on Unix domain socket “/var/run/postgresql/.s.PGSQL.5432"?

(Also I can't figure out how to run 'rake db:rollback' on the migration on Google Cloud SQL, but I don't think that's the problem.)

I'll try a few things things ...

.gitignore
mattlindsey: actually looks like I just accidentally removed the last blank line, which doesn't seem to hurt anything

Gemfile
mattlindsey: I took it out.

config/application.rb
mattlindsey: I took it out.

config/storage.yml
mattlindsey: I took it out.

db/migrate/20221221225406_create_active_storage_tables.active_storage.rb
mattlindsey: I took it out.

mattlindsey:
@mtc2013.  See it this looks good now.  We should squash the commits when merging since there's so many.

mtc2013 (approved):
Let's pull this in and open a ticket for streamlining the deploy to google cloud. It's not a showstopper since we're not doing continuous delivery right now, nor do we have a production site.

PR #3825, opened 2022-12-28
Improve projects page

### LINKED ISSUE #3827: Improve project list page
The current project page contains large images which are often not used and don't really add any benefit.
Also it has infinite scrolling, which is pretty useless since it's better to maintain a small set of active projects (I think).

Thus we should change it to a regular list, still allowing for filtering by language but also adding the ability to display 'inactive' project so (for example) one can see past/completed projects.

mattlindsey:
Made the projects page a simple list of only 'active' projects, although inactive/closed projects can be viewed by selecting a checkbox.  Note that the search dropdown and checkbox match the members page, and I took out the unnecessary simple_form gem.  

fixes #3827 

Screenshot AFTER:
image

Screenshot BEFORE
image


app/controllers/projects_controller.rb
```diff
@@ -101,17 +100,29 @@ def set_project
...
-  def query_projects(status)
-    status ||= 'active'
-    @projects = Project.where(status: status)
-                       .order('last_github_update DESC NULLS LAST')
-                       .order('commit_count DESC NULLS LAST')
-                       .includes(:user)
+  def query_projects_alpha
+    status = params[:status]
+    @projects = if status == 'pending'
+                  Project.where(status: %w(pending)).order('title ASC')
```
mtc2013: Can we try seeing if status: status works here instead of this %w(pending) thing?

mattlindsey: Fixed, and refactored duplicate code in query_projects.

mtc2013 (approved):
Ok this works for now.  Maybe if we make the index view a little less complicated in the future via helpers or something, we can improve this with a future ticket.

mtc2013:
We are getting one failure on semaphore.  This seems related to VCR possibly, or just network failure? @mattlindsey

mtc2013:
I kicked it off again and it works now.  Must have been internet weather

PR #3833, opened 2023-01-28
Use chromium for testing

### LINKED ISSUE #3832: google chrome in docker does not work on new mac chips
The docker install fails on the newest macbooks because of the new m3 chip not being supported by google chrome.  But chromium works.

mattlindsey:
fixes #3832


mtc2013 (approved):
It still works on my x64 and since @mattlindsey has it working on m2, I'm going to approve.

Issue #3834, opened 2023-01-29
Issue using runSpec in docker


mtc2013:
On linux and windows using docker, this error comes up when using runSpec command argument to npx:

root@e0262ecdc520:/WebsiteOne# npx jasmine-browser-runner runSpecs
Jasmine server is running here: http://localhost:37771
Jasmine tests are here:         /WebsiteOne/spec/javascripts
Source files are here:          /WebsiteOne/public/assets
 The ChromeDriver could not be found on the current PATH, trying Selenium Manager
/WebsiteOne/node_modules/selenium-webdriver/bin/linux/selenium-manager: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /WebsiteOne/node_modules/selenium-webdriver/bin/linux/selenium-manager)
Unable to obtain driver using Selenium Manager: Error: Error executing command with /WebsiteOne/node_modules/selenium-webdriver/bin/linux/selenium-manager,--browser,chrome
/WebsiteOne/node_modules/selenium-webdriver/bin/linux/selenium-manager: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /WebsiteOne/node_modules/selenium-webdriver/bin/linux/selenium-manager)
Error: The ChromeDriver could not be found on the current PATH.
      Please download the latest version of the ChromeDriver
      from http://chromedriver.storage.googleapis.com/index.html
      and ensure it can be found on your PATH.
    at new ServiceBuilder (/WebsiteOne/node_modules/selenium-webdriver/chrome.js:175:13)
    at Function.getDefaultService (/WebsiteOne/node_modules/selenium-webdriver/chrome.js:264:12)
    at Function.createSession (/WebsiteOne/node_modules/selenium-webdriver/chromium.js:669:49)
    at Function.createSession (/WebsiteOne/node_modules/selenium-webdriver/chrome.js:255:13)
    at createDriver (/WebsiteOne/node_modules/selenium-webdriver/index.js:147:33)
    at Builder.build (/WebsiteOne/node_modules/selenium-webdriver/index.js:675:16)
    at buildWebdriver (/WebsiteOne/node_modules/jasmine-browser-runner/lib/webdriver.js:33:55)
    at Object.runSpecs (/WebsiteOne/node_modules/jasmine-browser-runner/index.js:89:25)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
[...]

PR #3836, opened 2023-01-29
Fix bug #3834 runSpec argument to use headless

mtc2013:
Fix for https://github.com/AgileVentures/WebsiteOne/issues/3834. This seems to work for me now when I run:

npx jasmine-browser-runner runSpec


mattlindsey (approved):
This works.

PR #3837, opened 2023-01-31
Upgrade to rails 7

### LINKED ISSUE #3819: Upgrade to rails 7

Florence-Njeri:
This PR contains the Rails 7 upgrade
#### Issue addressed

fixes #3819


mattlindsey:
I think you can probably go right to version 7.0.4.2

mattlindsey:
Hi @Florence-Njeri.  I was bored and was looking at this and found a fix to many of the cucumber errors you are getting.

If you modify the file initializers/cucumber.rb to contain the following, many errors will disappear.  See change to line 11.

```
# frozen_string_literal: true

# Based on https://github.com/cucumber/cucumber-ruby/issues/1432
# HACK: this method was available in cucumber 3.1 and VCR relies on it to
# generate cassette names.
if Rails.env.test?
  Cucumber::Core::Test::Case.class_eval do
    def feature
      string = File.read(location.file)
      document = ::Gherkin::Parser.new.parse(string)
      document.feature
    end
  end
end
```

mattlindsey:
The problem above is referenced here:
https://github.com/cucumber/cucumber-ruby/issues/1432

mattlindsey:
I'm also thinking you can include the upgrade to ruby 3.2.1 if you want, if it doesn't have any issues with websiteone.  Or we can do a separate issue.
https://www.ruby-lang.org/en/news/2023/02/08/ruby-3-2-1-released/

Florence-Njeri:
Hi @mattlindsey, sure I'll include the Ruby upgrade too in this one too and also check the cucumber fix

mattlindsey:
I was playing with this and found a fix to some of the cucumber errors you are still getting.  Apparently 'to_date' and 'to_time' below are unnecessary and now cause errors.

Update lines 273 and 274 of 'features/step_definitions/event_steps.rb' from:
```
  expect(@start_date).to eq @tz.utc_to_local(@event.start_datetime).to_date.strftime('%Y-%m-%d')
  expect(@start_time).to eq @tz.utc_to_local(@event.start_datetime).to_time.strftime('%I:%M %p')
```
To:
```
  expect(@start_date).to eq @tz.utc_to_local(@event.start_datetime).strftime('%Y-%m-%d')
  expect(@start_time).to eq @tz.utc_to_local(@event.start_datetime).strftime('%I:%M %p')
``` 

Also lines 285 and 286 of 'features/step_definitions/event_steps.rb'

mattlindsey:
Replace code starting at line 47 of the same file to eliminate another error:
```
Given('following events exist:') do |table|
  table.hashes.each do |hash|
    hash[:project_id] = Project.find_by(title: hash['project']).id unless hash['project'].blank?
    hash[:start_datetime] = Time.current.strftime('%c') if hash[:start_datetime] == 'TODAYS_DATE'
    hash.delete('project')
    hash[:repeat_ends] = false
    create(:event, hash)
  end
end
```

Florence-Njeri:
> Replace code starting at line 47 of the same file to eliminate another error:
> 
> ```
> Given('following events exist:') do |table|
>   table.hashes.each do |hash|
>     hash[:project_id] = Project.find_by(title: hash['project']).id unless hash['project'].blank?
>     hash[:start_datetime] = Time.current.strftime('%c') if hash[:start_datetime] == 'TODAYS_DATE'
>     hash.delete('project')
>     hash[:repeat_ends] = false
>     create(:event, hash)
>   end
> end
> ```

Thank you

mattlindsey:
Hi @Florence-Njeri.  This is looking good, but unfortunately I think that I found another issue.  We need to remove the 'compass-rails' gem from Gemfile because it is very old and discontinued, and doesn't work with latest ruby.  It gives an error on the file.exists method, so I think it needs to be taken out.  (I tried to find a fix for it, but couldn't.)

It is a CSS library, so we also need to remove where it is included in app/assets/stylesheets/application.sccs, and fix where it is used in the styles -- which might be a lot of places.  I found this, which may guide us on what to fix:  https://medium.com/@ayushya/migrating-from-compass-to-node-sass-4dadae9e8fb7.  And might need to add this: https://github.com/thoughtbot/bourbon.

Do you think you can try to do this?  Do you have any concerns?  I can help if you want.
Another option you can try is to not upgrade ruby, since compass-rails may still work with 3.0.5.  (Actually that may be a better option.  Then we can create a different Issue for removing compass-rails.)

mattlindsey:
Hi @Florence-Njeri.  Actually it turns out we are not using much of compass-rails, so it was very easy to remove.  Just:
- remove from Gemfile
- remove from first line in app/assets/stylesheets/application.scss
- remove the few uses of 'border-radius' from app/assets/stylesheets/layout/navbar.scss

Florence-Njeri:
> Hi @Florence-Njeri. Actually it turns out we are not using much of compass-rails, so it was very easy to remove. Just:
> 
> * remove from Gemfile
> * remove from first line in app/assets/stylesheets/application.scss
> * remove the few uses of 'border-radius' from app/assets/stylesheets/layout/navbar.scss

Hi @mattlindsey the error is still present

mattlindsey:
> 

Hi @Florence-Njeri.  Which error?  The only error left that I see is when running this: `bundle exec rspec ./spec/models/event_instance_spec.rb`

mattlindsey:
Looks like the 'coveralls' gem needs to be updated to this: `gem 'coveralls_reborn', require: false`

See: https://docs.coveralls.io/ruby-and-rails

Florence-Njeri:
> Looks like the 'coveralls' gem needs to be updated to this: `gem 'coveralls_reborn', require: false`
> 
> See: https://docs.coveralls.io/ruby-and-rails

Let me take a look

Due to the error on Semphore:
```
Coveralls encountered an exception:01:03
NoMethodError01:03
undefined method `exists?' for File:Class01:03
/home/semaphore/WebsiteOne/vendor/bundle/ruby/3.2.0/gems/coveralls-0.7.1/lib/coveralls/configuration.rb:72:in `yaml_config'
```

mattlindsey:
Looks like 'Time.mktime' need to be format:   Time.mktime(year, month, day, hour, min, sec_with_frac)

See: https://www.rubydoc.info/stdlib/core/Time.mktime

Gemfile
```diff
@@ -26,7 +26,7 @@ gem 'cocoon'
 gem 'code_climate_badges', git: 'https://github.com/AgileVentures/codeclimate_badges'
 gem 'coderay'
 gem 'colored'
-gem 'compass-rails', '~> 4.0'
+# gem 'compass-rails', '~> 4.0'
```
mattlindsey: This line can be deleted.

app/assets/stylesheets/application.scss
```diff
@@ -1,4 +1,4 @@
-@import "compass/css3";
+// @import "compass/css3";
```
mattlindsey: This line can be deleted.

Gemfile
```diff
@@ -46,7 +46,7 @@ gem 'kaminari'
 gem 'kramdown', '~> 2.1'
 gem 'local_time', '~> 2.1'
 gem 'mercury-rails', git: 'https://github.com/AgileVentures/mercury.git'
-gem 'nokogiri', '1.12.0'
+gem 'nokogiri', '1.14.2'
```
mattlindsey: I think we want: `gem 'nokogiri', '~> 1.14.2'`

Gemfile
```diff
@@ -82,7 +82,7 @@ gem 'utf8-cleaner'
 gem 'vanity'
 gem 'verbs'
 gem 'will_paginate-bootstrap'
-gem 'youtube_rails'
+gem 'youtube_rails', '~>1.2.3'
```
mattlindsey: Put a space after the ~>

mattlindsey (changes requested):

Dockerfile
```diff
@@ -4,7 +4,7 @@
 # [START cloudrun_rails_base_image]
 # Pinning the OS to buster because the nodejs install script is buster-specific.
 # Be sure to update the nodejs install command if the base image OS is updated.
-FROM ruby:3.0-buster as base
+FROM ruby:3.2.1-buster as base
```
mattlindsey: Can you change this to: `FROM ruby:3.2.1 as base`
since the 'buster' version gives an error for me

mattlindsey (changes requested):

mattlindsey (approved):
Looks good!

PR #3838, opened 2023-01-31
add last_github_update to project list

### LINKED ISSUE #3835: Add last_github_update to Projects page
I think it would be helpful to add last_github_update on the page of Projects By Activity.

It's ordered that way already:
    query_projects('last_github_update DESC NULLS LAST')

mattlindsey:
fixes #3835 

Pic below shows the 'tool tip' when you hover over the date:

image

### DIFF
--- a/app/views/projects/index.html.erb
+++ b/app/views/projects/index.html.erb
@@ -44,6 +44,7 @@
     <% show_it = (p.status == 'active' || p.status == 'Active' ? true : false) %>
     <% follower_count = p.followers.count %>
     <% commit_count = p.commit_count %>
+    <% last_update = p.last_github_update %>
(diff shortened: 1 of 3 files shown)


app/views/projects/index.html.erb
```diff
@@ -44,6 +44,7 @@
     <% show_it = (p.status == 'active' || p.status == 'Active' ? true : false) %>
     <% follower_count = p.followers.count %>
     <% commit_count = p.commit_count %>
+    <% last_update = p.last_github_update %>
```
mtc2013, line 47: We should put in a ticket to take some of this logic and local variable declarations out of this view and put them in helpers or partials

mtc2013: To facilitate this, we should have some kind of test of this feature at the cuke level, at least.

mtc2013 (approved):
We're good.

PR #3843, opened 2023-02-13
Add current month past events to full calendar

### LINKED ISSUE #3842: Post events on calendar as they happen
Post events to the monthly calendar in a different color (to differentiate them) so that users can have a better overview of how many and which events actually occur, and perhaps click on them to show details.

Here is the PR that added the calendar for reference:
https://github.com/AgileVentures/WebsiteOne/pull/1937

mattlindsey:
fixes #3842 

Screenshot below shows 2 events before the current date:
image

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -21,10 +21,11 @@ class EventsController < ApplicationController
 
   def index
     @projects = Project.active
+    @events = Event.upcoming_events(specified_project)
     respond_to do |format|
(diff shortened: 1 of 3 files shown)


app/controllers/events_controller.rb
```diff
@@ -21,10 +21,12 @@ def show
 
   def index
     @projects = Project.active
-    @events = Event.upcoming_events(specified_project)
```
mtc2013, line 24: do we need to delete this?

mtc2013 (approved):
Approved

PR #3856, opened 2023-03-06
Add courses seed data

### LINKED ISSUE #3855: E-learning: Add seed data for Courses
30.times do
  Course.create!([{
    title: Faker::Educator.course_name,
    description: Faker::TvShows::GameOfThrones.quote
  }])
end

mattlindsey:
Add seed data for new Courses table/model.

fixes #3855


db/seeds.rb
```diff
@@ -77,6 +77,20 @@ def get_country
     )
 
     puts 'Created default projects'
+
+    opensource = u.courses.create!(
```
faithngetich: I see we have an object called opensource but I can't seem to find where this is referenced

db/seeds.rb
```diff
@@ -77,6 +77,20 @@ def get_country
...
+      title: 'OpenSource',
+      description: 'How to contribute to open source projects.',
+      status: 'Active'
+    )
+
+    gitcourse = u.courses.create!(
```
faithngetich: same as above. If we are not using those variables: `gitcourse` and `opensource`, we should consider removing them.

db/seeds.rb
mattlindsey: Fixed.

db/seeds.rb
mattlindsey: Fixed.

PR #3903, opened 2023-08-13
Added account confirmation emails for email registrations "fixes #3896"

### LINKED ISSUE #3896: Add account confirmation emails for email registrations.
Registration by email is currently disabled in production because otherwise hundreds of fake/spam accounts get generated monthly.

Add the requirement to 'confirm' email accounts by clicking on a link in a confirmation email.  The devise gem should have this feature.   Then we can re-enable email account registration on production (in production.yaml).

wordhouse66:
Using devise :confirmable I added email confirmation 

Feature spec in features/users/confirm_user_email.feature


.gitignore
```diff
@@ -102,8 +102,12 @@ certbot.log
 rerun.txt
 
 # Ignore VCR
+<<<<<<< HEAD
```
mattlindsey: These lines with "<<<" and "===" mean that there was a merge conflict in your changes.
I don't think this file needs to be modified for this PR, so can you restore it?

README.md
```diff
@@ -39,21 +39,17 @@ Later Raoul Diffou joined to take over as project manager as Thomas and Bryan ha
...
-rake mailer:send_welcome_message
-rake modify_event_participation
-rake paypal:create_paypal_plans
-rake stats
-rake user:create_anonymous
-rake vcr_billy_caches:reset
-```
-
-Updating the pages requires the administrator to run the `rake fetch_github:content_for_static_pages` task.
+## Walkthroughs
```
mattlindsey: Is this change from an old version of this file?  I don't think this file needs to be changed for this PR.

db/migrate/20230812061445_remove_reconfirmable_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveReconfirmableFromUsers < ActiveRecord::Migration[7.0]
```
mattlindsey: If this column isn't needed, you can delete the migration that adds it and the one that removes it.

